PR #18, opened 2020-09-01
Add CBMC Batch proof markers

karkhaz:
This adds a marker file to each of the proof directories so that they will be noticed by the proof CI.


aggarw13:
Will there be a separate step/PR for enabling proof checks with the old CI infrastructure in the repository?

karkhaz:
@aggarw13 nothing on your end, thank you! We'll enable CI once it's ready, hopefully tomorrow. There are a couple of steps we still need to take, but it'd EoD for us so we'll resume tomorrow. Thank you!

PR #20, opened 2020-09-01
Fix doxygen build due to renamed files

muneebahmed10:
*Description*:
Renaming files with the `core_` prefix broke doxygen builds as doxygen `@file` targets were not updated. This fixes the errors in the doxygen build

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,5 +58,5 @@ jobs:
           sudo apt-get install -y libclang-9-dev
       - name: Run Doxygen And Verify Stdout Is Empty
         run: |
-          doxygen docs/doxygen/config.doxyfile | tee doxyoutput.txt
+          doxygen docs/doxygen/config.doxyfile 2>&1 | tee doxyoutput.txt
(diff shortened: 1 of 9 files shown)


aggarw13:
Is the doxygen job in GitHub Actions insufficient to not catch the build errors?

muneebahmed10:
> Is the doxygen job in GitHub Actions insufficient to not catch the build errors?

Correct, doxygen errors are written to `stderr`, and I think the job only checkout `stdout`. This commit https://github.com/FreeRTOS/coreMQTT/pull/6/commits/e4827bf009e8e410a2bea9c5c7e9093ff75a8f15 was added to fix it

aggarw13:
> > Is the doxygen job in GitHub Actions insufficient to not catch the build errors?
> 
> Correct, doxygen errors are written to `stderr`, and I think the job only checkout `stdout`. This commit [e4827bf](https://github.com/FreeRTOS/coreMQTT/commit/e4827bf009e8e410a2bea9c5c7e9093ff75a8f15) in #6 was added to fix it

Can you pull that commit within this PR so that we can have the confidence that these fixes suffice?

PR #22, opened 2020-09-01
Add config to allow MQTT build without custom config

aggarw13:
### Problem

Currently, the MQTT library requires `core_mqtt_﻿config.h` to be provided by application to build the library (i.e. library has a **hard-dependency** on custom config file). 
However, customers may want to build the library by just using the default values present in `core_mqtt_config_defaults.h` instead of providing their custom values.

### Solution

Remove the hard-dependency of library on custom config by provide a `mqttDO_NOT_USE_CUSTOM_CONFIG` build config to allow customers to build MQTT library without explicitly providing a custom config (i.e. build with default values from `core_mqtt_config_defaults.h`).
Note: **By default**, the MQTT library _will_ require a custom config for building. On encountering compiler error, this will _make the user look_ at the available configuration options, and decide whether the default values suit their platform/application needs.


source/include/core_mqtt.h
```diff
@@ -26,11 +26,16 @@
...
 
-/* Include main config file before other headers. */
-#include "core_mqtt_config.h"
-
-/* Include config defaults file after main config file to get
- * default values of configs not defined in main config file. */
+/* mqttDO_NOT_USE_CUSTOM_CONFIG allows building the MQTT library
+ * without a custom config. If a custom config is provided, the
+ * mqttDO_NOT_USE_CUSTOM_CONFIG macro should not be defined. */
+#ifndef mqttDO_NOT_USE_CUSTOM_CONFIG
```
sarenameas, line 32: Why is mqtt lowercase? This is different than other macros in the library.

README.md
```diff
@@ -4,7 +4,17 @@ This repository contains an ISO C90 compliant MQTT client library designed for e
...
-### Prerequisites
+### MQTT Config File
+
+The MQTT client library exposes build configuration macros that are required for building the library.
+A list of all the configurations and their default values are defined in [core_mqtt_config_defaults.h](https://github.com/FreeRTOS/coreMQTT/blob/master/source/include/core_mqtt_config_defaults.h). 
+To provide custom values for the configuration macros, a custom config file named `core_mqtt_config.h` can be
+provided to the library.
+
+By default, the `core_mqtt_config.h` custom config is required to build the library. To disable the requirement
+and build the library with the default configuration values, provide the `mqttDO_NOT_USE_CUSTOM_CONFIG` as a compile time preprocessor macro.
```
sarenameas, line 15: `mqttDO_NOT_USE_CUSTOM_CONFIG` is a proper noun and does need a "the" before it
```suggestion
and build the library with the default configuration values, provide `mqttDO_NOT_USE_CUSTOM_CONFIG` as a compile time preprocessor macro.
```

README.md
```diff
@@ -4,7 +4,17 @@ This repository contains an ISO C90 compliant MQTT client library designed for e
...
 
-### Prerequisites
+### MQTT Config File
+
+The MQTT client library exposes build configuration macros that are required for building the library.
+A list of all the configurations and their default values are defined in [core_mqtt_config_defaults.h](https://github.com/FreeRTOS/coreMQTT/blob/master/source/include/core_mqtt_config_defaults.h). 
+To provide custom values for the configuration macros, a custom config file named `core_mqtt_config.h` can be
+provided to the library.
+
+By default, the `core_mqtt_config.h` custom config is required to build the library. To disable the requirement
```
sarenameas, line 14:
```suggestion
By default, the `core_mqtt_config.h` custom config is required to build the library. To disable this requirement
```

source/include/core_mqtt.h
aggarw13: Sure, changed it. Was just biased by the naming pattern here: https://github.com/aws/amazon-freertos/blob/v1.4.8/lib/include/aws_mqtt_lib.h#L38

source/include/core_mqtt_lightweight.h
```diff
@@ -46,11 +46,16 @@
...
+ * MQTT_DO_NOT_USE_CUSTOM_CONFIG macro should not be defined. */
+#ifndef MQTT_DO_NOT_USE_CUSTOM_CONFIG
+    /* Include custom config file before other headers. */
+    #include "core_mqtt_config.h"
+#endif
 
-/* Include config defaults file after main config file to get
- * default values of configs not defined in main config file. */
+/* Include config defaults header to get default values of configs not
+ * defined in custom main config file. */
```
sarenameas, line 58: May i suggest? 
```suggestion
 * defined in core_mqtt_config.h. */
```

aggarw13: Fixed

sarenameas (dismissed):

source/include/core_mqtt.h
RichardBarry: I'm not sure why this is necessary.  Can the user just leave the config file empty if they want to use the default values?

aggarw13: We discussed about it internally, and thought that it would be good to not force the customer to define a config file if they don't want to define custom values. 
We decided to mimic the approach that was followed in a previous MQTT release: https://github.com/aws/amazon-freertos/blob/release-1.4.x/lib/include/aws_mqtt_lib.h#L38

cobusve (approved):

abhidixi11 (approved):

source/include/core_mqtt.h
aggarw13: We discussed about it separately, and agreed that this is the best option we have of all options considered for customer usability experience. The concern of ambiguity for the customer is mitigated with clear information on the REAMDE about the options of building the library with or without config file

PR #23, opened 2020-09-02
Address MISRA 2.3 Rule violation

aggarw13:
Replace MQTTSubAckStatus_t enum with equivalent macros to address MISRA 2.3 Rule violation﻿


aggarw13:
We have decided to not make the change from **enum** to **macro constants** to suppress the MISRA violation as using enum type is preferred

PR #35, opened 2020-09-04
Create LICENSE.md

muneebahmed10:
Adds MIT license, copyright holder Amazon Web Services

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020 Amazon Web Services
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


LICENSE.md
```diff
@@ -0,0 +1,21 @@
+MIT License
+
```
aggarw13, line 2: Is this the same as the C-SDK repository?

muneebahmed10: Yes, it's identical

aggarw13 (approved):

sarenameas:
Looks like we can close this?

muneebahmed10:
Looks like this file was added in another commit without a PR. Unsure if we still need to add a copyright holder but that would be a separate PR

PR #28, opened 2020-09-02
Remove library target, and move CMake files under test folder

aggarw13:
Remove presence of CMake build infrastructure in the root folder by moving the CMakeLists.txt file to `test` folder﻿
Also, remove the `core_mqtt` library target

PR #37, opened 2020-09-04
Re-adding MQTT library target to test MQTT_DO_NOT_USE_CUSTOM_CONFIG

aggarw13:
Support Coverity analysis on MQTT files (for static analysis and MISRA compliance check) by re-adding the `core_mqtt` library (that was removed in #28)
Presence of `core_mqtt` target also provides mechanism of testing the `MQTT_DO_NOT_USE_CUSTOM_CONFIG` flag (introduced in #22)﻿


yourslab (dismissed):

yourslab (approved):

README.md
```diff
@@ -21,21 +21,19 @@ and build the library with default configuration values, provide `MQTT_DO_NOT_US
 
 ### Platform Prerequisites
```
yourslab, line 22: Why does the `Build Unit Tests.` have a period at the end? Also, isn't it better to have the reference examples before the build instructions for unit tests?

test/CMakeLists.txt
```diff
@@ -0,0 +1,90 @@
+cmake_minimum_required ( VERSION 3.13.0 )
```
sarenameas, line 1: Why are we moving out to the test/ folder?

test/CMakeLists.txt
```diff
@@ -0,0 +1,90 @@
...
+set( CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib )
+set( CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib )
+
+# ===================================== Library Configuration =================================================
+
+# Include filepaths for source and include.
+include( ${MODULE_ROOT_DIR}/mqttFilePaths.cmake )
+
+# MQTT library target.
+add_library( core_mqtt
```
sarenameas, line 38: I have spoken with Nathan and sent him all of the commands to run coverity using the unit test target as it is. The only thing that would use extra target is coverity, so I do not think an extra target is necessary. This extra target does not help with the advisory 8.7 and 2.3 rules.

sarenameas (dismissed):

test/CMakeLists.txt
aggarw13: The extra target enables us to test the `MQTT_DO_NOT_USE_CUSTOM_CONFIG` macro, which cannot be tested through the unit test target which requires use of a `core_mqtt_config.h` file to provide definition of `NetworkContext_t`

test/CMakeLists.txt
aggarw13: This PR adds a mechanism for testing the `MQTT_DO_NOT_USE_CUSTOM_CONFIG` macro (that allows building the MQTT library _as-is_ without a custom config) through the library target.

test/CMakeLists.txt
sarenameas: Ok I can't think of any other way to test other than the CSDK demos

aggarg (approved):

aggarg (approved):

yourslab (approved):

PR #39, opened 2020-09-08
Use Litani for CBMC proofs on master branch

karkhaz:
This PR contains the commits that allow CBMC proofs to run using Litani. We have been running CI from the litani-mainline branch where these commits are stored, but are now proposing to merge these commits into master.


karkhaz:
The "CBMC Batch" jobs are expected to fail on this PR. The PR is not to be merged until we switch off CBMC Batch.

.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "test/cbmc/aws-templates-for-cbmc-proofs"]
 	path = test/cbmc/aws-templates-for-cbmc-proofs
 	url = https://github.com/awslabs/aws-templates-for-cbmc-proofs.git
+[submodule "test/cbmc/litani"]
```
markrtuttle, line 7: I recommend that this be submoduled into the tools directory parallel to the template starter kit.

test/cbmc/proofs/MQTT_Connect/MQTT_Connect_harness.c
```diff
@@ -26,7 +26,7 @@
 #include "core_mqtt.h"
 #include "mqtt_cbmc_state.h"
 
-void harness()
+void MQTT_Connect_harness()
```
markrtuttle, line 29: Why change the harness names?   This is inconsistent with the starter kit templates in the master branch of aws-templates-for-cbmc-proofs.

test/cbmc/proofs/Makefile-project-defines
```diff
@@ -14,6 +14,11 @@
 COMPILE_FLAGS += -fPIC
 COMPILE_FLAGS += -std=gnu90
 
+# Path to litani executable, used for running proofs and displaying report
+LITANI ?= $(SRCDIR)/test/cbmc/litani/litani
```
markrtuttle, line 18: This path would have to change if litani is installed in the tools directory

test/cbmc/proofs/run-cbmc-proofs.py
```diff
@@ -0,0 +1,269 @@
...
+    else:
+        proofs_remaining = []
+
+    for root, _, fyles in os.walk(proof_root):
+        proof_name = str(pathlib.Path(root).name)
+        if proof_list and proof_name not in proof_list:
+            continue
+        if proof_list and proof_name in proofs_remaining:
+            proofs_remaining.remove(proof_name)
+        if "cbmc-batch.yaml" in fyles:
```
markrtuttle, line 135: Are we really sticking with cbmc-batch.yaml as the marker?  I thought you had a different scheme for after the switch to litani.

test/cbmc/proofs/run-cbmc-proofs.py
```diff
@@ -0,0 +1,269 @@
+#!/usr/bin/env python3
```
markrtuttle, line 1: Is this just the generic run-cbmc-proofs script?  Is there anything special about this version?  Can you comment and hightlight it so even I can find it?

markrtuttle (changes requested):
This looks good and straightforward, but changing the name of the harness in the proof harness and changing the naming schemes in the individual proof Makefiles is inconsistent with the master branch of the starter kit templates.

test/cbmc/proofs/run-cbmc-proofs.py
karkhaz: Good idea, thank you! I have commented on the changes, starting with the string `# PROJECT SPECIFIC`, and will start to do this for all the other projects also. That should be very helpful if we ever decide to make a unified script that gets added to the starter kit!

In summary, this script differs because it is setting the Kissat environment variable when invoking Make. So there's some checking around that. The user can use the native solver (minisat) if they like, by passing an explicit command-line flag.

test/cbmc/proofs/run-cbmc-proofs.py
karkhaz: We must, because we're running Batch in parallel with the new CI. After Batch is retired, it will be possible to rename these files in a PR. This won't require any modification to the new CI; you can do everything you need with just a PR.

.gitmodules
karkhaz: The template starter kit currently resides under `test/cbmc`, not under the tools directory. So I did put litani parallel to the starter kit. I think you're thinking of the old repo?

karkhaz: Or are you suggesting that I move both of the submodules to he `tools` directory in this PR?

test/cbmc/proofs/MQTT_Connect/MQTT_Connect_harness.c
karkhaz: This is how all the other projects do it...s2n makes the name of the harness the same as the name of the file.

The real reason: I would like every proof to have a "name" for litani to refer to it by. (The argument to the `--pipeline-name` flag). So in the starter kit, the pipeline name is set to `$(HARNESS_ENTRY)`. That's how the table looks [like this](https://d2s5j4r0gkya4b.cloudfront.net/5d596477-4746-4cb2-9b6e-4a66621b5616/index.html), the rows are the harness entry names.

So: currently we have to have it this way, because that's what's in Makefile.common. If you don't like this, we should change it in the starter kit, but I recommend that we keep it this way in MQTT for now. Suggest we discuss this offline.

markrtuttle: You misunderstand the starter kit.  The Makefile under template-for-proof specifies:

```

Comment (markrtuttle) [reply test/cbmc/proofs/MQTT_Connect/MQTT_Connect_harness.c:29]: Furthermore, we are doing out of source builds, so there is a unique pathname referring to the proof harness source code itself.

File: .gitmodules (hunk 1 above)
Comment (markrtuttle) [reply .gitmodules:7]: No, it should be parallel to the starter kit, so you might be doing the right thing.

File: test/cbmc/proofs/MQTT_Connect/MQTT_Connect_harness.c (hunk 2 above)
Comment (karkhaz) [reply test/cbmc/proofs/MQTT_Connect/MQTT_Connect_harness.c]: I've reverted this change as discussed offline, and introduced a new variable for the unique name of the proof. Thank you!

File: test/cbmc/proofs/run-cbmc-proofs.py [hunk 6]
```diff
@@ -0,0 +1,278 @@
...
+            "help": "Project name for report. Default: %(default)s",
+    }, {
+
+    # PROJECT SPECIFIC
+    # This flag exists because we wish to use an external SAT solver for this
+    # project by default, both locally and in CI. This flag is provided to make
+    # it easy to invoke the native solver if needed.
+            "flags": ["--use-native-solver"],
+            "action": "store_true",
+            "help": "Use CBMC's native SAT solver instead of kissat (proofs will run for longer)",
```
markrtuttle, line 94: This --use-native-solver is an argument to the script.  Does the new CI invoke this script with this flag?  I expected to see this flag in the CBMCFLAGS definition in the proof Makefiles for the slow proofs needing this solver, but I don't see that here.

markrtuttle (dismissed):

test/cbmc/proofs/run-cbmc-proofs.py
karkhaz: The new CI does not use this flag. Both the CI and the developer are expected to *not* use this flag. It's provided just in case you'd like to run minisat.

What this is doing is running _all_ proofs with Kissat.

PR #42, opened 2020-09-09
Add documentation for MQTT_DO_NOT_USE_CUSTOM_CONFIG

sarenameas:
MQTT_DO_NOT_USE_CUSTOM_CONFIG needed to be documented in the Doxygen configs page.
The Doxygen configs page shows all the possible configurations of the library.

image


docs/doxygen/pages.txt
```diff
@@ -71,6 +71,9 @@ Some configuration settings are C pre-processor constants, and some are function
 @section MQTT_MAX_CONNACK_RECEIVE_RETRY_COUNT
 @copydoc MQTT_MAX_CONNACK_RECEIVE_RETRY_COUNT
 
+@section MQTT_DO_NOT_USE_CUSTOM_CONFIG
```
aggarw13, line 74: IMO, it would be better to have this config at the top of the file to set the context about the importance of the custom config and the config defaults (from both the perspective of reading the `core_mqtt_config_defaults.h` file and the doxygen output)

sarenameas: sure

source/include/core_mqtt_config_defaults.h
```diff
@@ -170,4 +170,17 @@
...
 #endif
 
+/* This macro definition is for Doxygen documentation only. */
+
+/**
+ * @brief Define this macro to build the MQTT library without the custom config
+ * file core_mqtt_config.h.
+ *
+ * If a custom config is provided, then MQTT_DO_NOT_USE_CUSTOM_CONFIG should not
+ * be defined.
```
aggarw13, line 180:
```suggestion
 * @brief Define this macro to build the MQTT library without the custom config
 * file core_mqtt_config.h. Without the custom config, the MQTT library builds with
 * default values of config macros defined in core_mqtt_config_defaults.h file.
 *
 * If a custom config is provided, then MQTT_DO_NOT_USE_CUSTOM_CONFIG should not
 * be defined.
```

aggarw13 (approved):

dan4thewin (approved):

PR #43, opened 2020-09-09
Use spellcheck from the CSDK repo

sarenameas:
Depends on: https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/1212

We should not duplicate the spellcheck scripts and use the one in the CSDK repo instead.
This is an interim solution while we work on creating a separate repo for the spellcheck scripts.


aggarw13:
Instead of having this repositories CI reference one of the hub-repositories (like the C-SDK), why not have a mechanism (like GitHub Actions) to monitor the duplicity of all the spell-check scripts and flag us when they differ?

sarenameas:
> Instead of having this repositories CI reference one of the hub-repositories (like the C-SDK), why not have a mechanism (like GitHub Actions) to monitor the duplicity of all the spell-check scripts and flag us when they differ?

Because we are still repeating the scripts and they are buggy.

nateglims:
If we could get the repo approved, I think we should put it into it's own repo as an action. We could expose a cleaner API, and (if I understand the docs correctly) it would be cached and potentially faster than cloning.

leegeth (approved):

sarenameas:
> If we could get the repo approved, I think we should put it into it's own repo as an action. We could expose a cleaner API, and (if I understand the docs correctly) it would be cached and potentially faster than cloning.

A separate repo was guested by Dan, but there is a lot of process in that, so i proposed this solution next.

nateglims:
I think this is an appropriate interim solution. Better to stop the rot now.

dan4thewin (approved):

PR #56, opened 2020-09-12
Session present flag fix.

abhidixi11:
Set session present variable explicitly to false when session present flag is not set in CONN_ACK, ( in case application does not  initialize the value).


aggarw13:
Should there be an accompanying unit test update as well to make sure that the `sessionPresent` flag is set to `false` even if the initialized value in the test is `true` (when the CONNACK response doesn't have the session present bit set as false)

aggarw13 (approved):

nateglims (approved):

muneebahmed10 (approved):

leegeth (approved):

PR #71, opened 2020-09-18
Set NULL payloads and fix unit test

muneebahmed10:
*Description*:
Set payload to NULL for zero length payloads:
 * `MQTT_DeserializePublish` sets `pPayload` to point to the byte after the topic name (or packet ID for QoS > 0). If there is no payload, then this could point to garbage data in the buffer. `payloadLength` is correctly set to 0 for this case, but it also makes sense to set the payload to NULL as well.

Improve unit tests for MQTT_DeserializePublish:
 * The unit tests for `MQTT_DeserializePublish` don't reset its `publishInfo` struct between test cases. Before a test, its pointers point to constant strings, but are changed to point to the serialized buffer afterwards. This is an issue for the next test as it will try to serialize the data into the same buffer that is being pointed to, and mangles the data.
 * The deserialized publish information isn't checked, only the return code. The above test issue was caught when adding these checks.
 * Added unit test to check for a NULL pointer when payloads are zero length.

### DIFF
--- a/source/core_mqtt_serializer.c
+++ b/source/core_mqtt_serializer.c
@@ -1298,7 +1298,7 @@ static MQTTStatus_t deserializePublish( const MQTTPacketInfo_t * pIncomingPacket
         }
 
         /* Set payload if it exists. */
-        pPublishInfo->pPayload = ( pPublishInfo->payloadLength != 0 ) ? pPacketIdentifierHigh : NULL;
+        pPublishInfo->pPayload = ( pPublishInfo->payloadLength != 0U ) ? pPacketIdentifierHigh : NULL;
 
         LogDebug( ( "Payload length %lu.", ( unsigned long ) pPublishInfo->payloadLength ) );
     }
--- a/source/core_mqtt_serializer.c
+++ b/source/core_mqtt_serializer.c
@@ -1274,6 +1274,9 @@ static MQTTStatus_t deserializePublish( const MQTTPacketInfo_t * pIncomingPacket
 
             LogDebug( ( "Packet identifier %hu.", *pPacketId ) );
 
+            /* Advance pointer two bytes to start of payload as in the QoS 0 case. */
+            pPacketIdentifierHigh += sizeof( uint16_t );
+
             /* Packet identifier cannot be 0. */
             if( *pPacketId == 0U )
             {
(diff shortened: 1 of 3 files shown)


test/unit-test/core_mqtt_serializer_utest.c
```diff
@@ -1660,10 +1654,10 @@ void test_MQTT_DeserializePublish( void )
...
     /* Create a PUBLISH packet to test. */
     memset( &publishInfo, 0x00, sizeof( publishInfo ) );
-    publishInfo.pTopicName = "/test/topic";
-    publishInfo.topicNameLength = ( uint16_t ) strlen( publishInfo.pTopicName );
-    publishInfo.pPayload = "Hello World";
-    publishInfo.payloadLength = ( uint16_t ) strlen( publishInfo.pPayload );
+    publishInfo.pTopicName = TEST_TOPIC_NAME;
+    publishInfo.topicNameLength = TEST_TOPIC_NAME_LENGTH;
+    publishInfo.pPayload = MQTT_SAMPLE_PAYLOAD;
+    publishInfo.payloadLength = MQTT_SAMPLE_PAYLOAD_LEN;
```
yourslab, line 1660: Could you possible use `setupWillInfo` for these instead, changing what is necessary for the specific test case? Maybe using the built in unity fixture, `setUp`, can make this easier for you.

source/core_mqtt_serializer.c
```diff
@@ -1286,17 +1289,17 @@ static MQTTStatus_t deserializePublish( const MQTTPacketInfo_t * pIncomingPacket
...
-        }
-        else
+        pPublishInfo->payloadLength = pIncomingPacket->remainingLength - pPublishInfo->topicNameLength - sizeof( uint16_t );
+
+        if( pPublishInfo->qos != MQTTQoS0 )
         {
-            pPublishInfo->payloadLength = ( pIncomingPacket->remainingLength - pPublishInfo->topicNameLength - 2U * sizeof( uint16_t ) );
-            pPublishInfo->pPayload = pPacketIdentifierHigh + sizeof( uint16_t );
+            /* Two more bytes for the packet identifier. */
+            pPublishInfo->payloadLength -= sizeof( uint16_t );
```
yourslab, line 1297: Why is this subtracting?

yourslab, line 1297: Instead of adding two bytes in advance then subtracting if `pPublishInfo->qos != MQTTQoS0`, why not just add once if `pPublishInfo->qos != MQTTQoS0`?

test/unit-test/core_mqtt_serializer_utest.c
muneebahmed10: Sure, I'll rename it to `setupPublishInfo`

source/core_mqtt_serializer.c
muneebahmed10: Not sure what you mean, we're not adding any bytes to the payload length. The payload length is defined as: Remaining length - # of Topic length bytes - Topic length - # of packet ID bytes. There are two packet ID bytes for QoS > 0, and 0 none for QoS 0. So, we need to subtract an additional 2 for QoS > 0.

yourslab: Oh got it, I seem to have misread the code.

yourslab (approved):

test/unit-test/core_mqtt_serializer_utest.c
```diff
@@ -1688,8 +1679,14 @@ void test_MQTT_DeserializePublish( void )
     mqttPacketInfo.pRemainingData = &buffer[ 2 ];
     status = MQTT_DeserializePublish( &mqttPacketInfo, &packetIdentifier, &publishInfo );
     TEST_ASSERT_EQUAL_INT( MQTTSuccess, status );
+    TEST_ASSERT_EQUAL_INT( TEST_TOPIC_NAME_LENGTH, publishInfo.topicNameLength );
+    TEST_ASSERT_EQUAL_MEMORY( TEST_TOPIC_NAME, publishInfo.pTopicName, TEST_TOPIC_NAME_LENGTH );
+    TEST_ASSERT_EQUAL_INT( MQTT_SAMPLE_PAYLOAD_LEN, publishInfo.payloadLength );
+    TEST_ASSERT_EQUAL_MEMORY( MQTT_SAMPLE_PAYLOAD, publishInfo.pPayload, MQTT_SAMPLE_PAYLOAD_LEN );
 
     memset( ( void * ) &mqttPacketInfo, 0x00, sizeof( mqttPacketInfo ) );
```
n9wxu, line 1687: the cast to void * should not be required and looks strange to me.  The point of a void * is to be a permissive pointer that accepts whatever it gets.

n9wxu (approved):

test/unit-test/core_mqtt_serializer_utest.c
muneebahmed10: Removed

yourslab (approved):

PR #75, opened 2020-09-25
Update submodule configuration for hub repo cloning efficiency

aggarw13:
Reduce the cloning time of hub repositories (`aws/aws-iot-device-sdk-embedded-C`, `aws/amazon-freertos`) that consume this repository by updating submodules to not be auto-cloned for `git submodule update --init --recursive` command in the hub repositories

PR #76, opened 2020-09-28
Fix BUILD_CLONE_SUBMODULES CMake config option

aggarw13:
With submodule configuration changes made in #75, the `BUILD_CLONE_SUBMODULES` CMake config became disfynctional (as it uses the `git submodule update --recursive` command).

This PR updates the CMake logic to use `git submodule update --checkout --init` command to fix the CMake config option.
Also, this PR changes the default value of `BUILD_CLONE_SUBMODULES` to be `OFF` to avoid _always_ auto-cloning the **CMock** submodule.

### DIFF
--- a/test/unit-test/cmock_build.cmake
+++ b/test/unit-test/cmock_build.cmake
@@ -3,7 +3,7 @@ macro( clone_cmock )
         find_package( Git REQUIRED )
         message( "Cloning submodule CMock." )
         execute_process( COMMAND rm -rf ${CMOCK_DIR}
-                         COMMAND ${GIT_EXECUTABLE} submodule update checkout --init --recursive ${CMOCK_DIR}
+                         COMMAND ${GIT_EXECUTABLE} submodule update --checkout --init --recursive ${CMOCK_DIR}
                         WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
                         RESULT_VARIABLE CMOCK_CLONE_RESULT )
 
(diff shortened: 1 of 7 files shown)


yourslab (approved):

test/unit-test/cmock_build.cmake
```diff
@@ -3,7 +3,7 @@ macro( clone_cmock )
         find_package( Git REQUIRED )
         message( "Cloning submodule CMock." )
         execute_process( COMMAND rm -rf ${CMOCK_DIR}
-                         COMMAND ${GIT_EXECUTABLE} submodule update --init --recursive ${CMOCK_DIR}
+                         COMMAND ${GIT_EXECUTABLE} submodule update checkout --init --recursive ${CMOCK_DIR}
```
sarenameas, line 6: Does this need to be --checkout?

sarenameas: If we take out the clone you added to the ci.yml, we could test this?

aggarw13: >Does this need to be --checkout?

Yes, that is how the `update = none` option is overriden

aggarw13: > If we take out the clone you added to the ci.yml, we could test this?

Sure

sarenameas (approved):

aggarg (approved):

PR #83, opened 2020-10-06
Resend PUBACK and PUBREC irrespective of the DUP flag

yourslab:
In the OASIS MQTT Version 3.1.1 OASIS Standard, the following is stated:

> After it has sent a PUBACK Packet the Receiver MUST treat any incoming PUBLISH packet that contains the same Packet Identifier as being a new publication, irrespective of the setting of its DUP flag.

Also, when AWS IoT Core resends a PUBLISH with the same packet ID as the first that wasn't PUBACK'd, the DUP flag will always be set to 0. Therefore, our library will return `MQTTStateCollision` and will NOT send a PUBACK in response.

For QoS 2, the OASIS spec does not seem to explicitly state anything about whether or not to resend a PUBREC in response to a resent PUBLISH with the DUP flag set to 0. However, I think the same logic still applies, so the PUBREC must still be resent in that case. Feel free to disagree.

### DIFF
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -1069,7 +1069,7 @@ static MQTTStatus_t handleIncomingPublish( MQTTContext_t * pContext,
 
             if( publishInfo.dup == false )
             {
-                LogError( ( "DUP flag is 0 for duplicate publish packet." ) );
+                LogError( ( "DUP flag is 0 for duplicate incoming publish packet." ) );
             }
         }
         else
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -1069,7 +1069,7 @@ static MQTTStatus_t handleIncomingPublish( MQTTContext_t * pContext,
 
             if( publishInfo.dup == false )
             {
-                LogError( ( "DUP flag is 0 for duplicate incoming publish packet." ) );
+                LogError( ( "DUP flag is 0 for duplicate packet (MQTT-3.3.1.-1)." ) );
             }
         }
         else
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -1064,15 +1064,12 @@ static MQTTStatus_t handleIncomingPublish( MQTTContext_t * pContext,
             publishRecordState = MQTT_CalculateStatePublish( MQTT_RECEIVE,
                                                              publishInfo.qos );
 
-            if( publishInfo.dup )
+            LogDebug( ( "Incoming publish packet with packet id %u already exists.",
+                        packetIdentifier ) );
+
+            if( publishInfo.dup == false )
             {
-                LogDebug( ( "Incoming publish packet with packet id %u already exists.",
-                            packetIdentifier ) );
-            }
-            else
-            {
-                LogError( ( "Incoming publish packet with packet id %u already exists, but DUP flag is 0.",
-                            packetIdentifier ) );
+                LogError( ( "DUP flag is 0 for duplicate publish packet." ) );
             }
         }
         else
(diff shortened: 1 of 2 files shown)


muneebahmed10:
> After it has sent a PUBACK Packet the Receiver MUST treat any incoming PUBLISH packet that contains the same Packet Identifier as being a new publication, irrespective of the setting of its DUP flag.

We already do this, since `MQTTStateCollision` can't be returned after the puback has been sent (since it's removed from the record). If the AWS IoT broker doesn't set the dup flag, it sounds like that's the fault of the broker. However, I'm fine with making this change if it doesn't violate the MQTT spec

source/core_mqtt.c
```diff
@@ -1054,7 +1054,7 @@ static MQTTStatus_t handleIncomingPublish( MQTTContext_t * pContext,
          *       state engine. This will be handled by ignoring the
          *       #MQTTStateCollision status from the state engine. The publish
          *       data is not passed to the application. */
-        else if( ( status == MQTTStateCollision ) && ( publishInfo.dup == true ) )
+        else if( status == MQTTStateCollision )
```
muneebahmed10, line 1057: If we're making this change specifically for AWS IoT Core, we should at least log a warning, if not an error. If the DUP flag isn't set, it seems to be a protocol violation ([3.3.1.1](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718038)): 
> The DUP flag MUST be set to 1 by the Client or Server when it attempts to re-deliver a PUBLISH Packet

yourslab: Please see the updated logs.

source/core_mqtt.c
```diff
@@ -1063,8 +1063,14 @@ static MQTTStatus_t handleIncomingPublish( MQTTContext_t * pContext,
              * for the duplicate incoming publish. */
             publishRecordState = MQTT_CalculateStatePublish( MQTT_RECEIVE,
                                                              publishInfo.qos );
+
             LogDebug( ( "Incoming publish packet with packet id %u already exists.",
                         packetIdentifier ) );
+
+            if( publishInfo.dup == false )
+            {
+                LogError( ( "DUP flag is 0 for duplicate incoming publish packet." ) );
```
muneebahmed10, line 1072: Should we have a more ominous sounding error message like "Protocol violation: ..."? I think we should be clear that this isn't according to the spec.

muneebahmed10 (dismissed):

source/core_mqtt_serializer.c
```diff
@@ -955,7 +955,7 @@ static MQTTStatus_t deserializeConnack( const MQTTPacketInfo_t * pConnack,
         if( ( pRemainingData[ 0 ] & MQTT_PACKET_CONNACK_SESSION_PRESENT_MASK )
             == MQTT_PACKET_CONNACK_SESSION_PRESENT_MASK )
         {
-            LogWarn( ( "CONNACK session present bit set." ) );
+            LogInfo( ( "CONNACK session present bit set." ) );
```
muneebahmed10, line 958: Thanks. I'm pretty sure this was a leftover from v4_beta, where the session present flag wasn't given back to the application

muneebahmed10 (approved):

source/core_mqtt.c
cobusve: I would simply say "DUP flag is 0 for duplicate packet (MQTT-3.3.1.-1)" that gives all the breadcrumbs you need to figure it out. If you just google MQTT-3.3.1.-1 you will hit it directly.

muneebahmed10 (approved):

cobusve (approved):

PR #89, opened 2020-10-22
Add more detailed comment about error code in transport interface

yourslab:
Add more detailed comment about error code in transport interface

### DIFF
--- a/source/interface/transport_interface.h
+++ b/source/interface/transport_interface.h
@@ -82,7 +82,13 @@
  * @snippet this define_transportrecv
  * <br>
  * This function is expected to populate a buffer, with bytes received from the
- * transport, and return the number of bytes placed in the buffer.
+ * transport, and return the number of bytes placed in the buffer. However, if a
+ * socket error occurs while trying to receive a message, then a negative error
+ * code will be returned. The specific error code depends on the implementation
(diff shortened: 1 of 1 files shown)


source/portable/transport_interface.h
```diff
@@ -82,7 +82,13 @@
  * @snippet this define_transportrecv
  * <br>
  * This function is expected to populate a buffer, with bytes received from the
- * transport, and return the number of bytes placed in the buffer.
+ * transport, and return the number of bytes placed in the buffer. However, if a
+ * socket error occurs while trying to receive a message, then a negative error
```
abhidixi11, line 86: Let's not say socket error, since underlying transport could be using something other than socket.

source/portable/transport_interface.h
```diff
@@ -82,7 +82,13 @@
...
+ * code will be returned. The specific error code depends on the implementation
+ * of the function used. It is also possible to have a return value that is less
+ * than the size of the payload such as when the TCP receive buffer is not large
```
abhidixi11, line 89: This condition will only happen when less number of bytes are received than the buffer size. We should not talk about TCP buffer. Let's just say '  It is also possible to have a return value that is less than the size requested '

PR #95, opened 2020-10-29
Align wording of Init with porting guide

muneebahmed10:
*Description*:
Change note in documentation of `MQTT_Init` to explicitly mention which functions need the time function as the porting guide does.


source/include/core_mqtt.h
```diff
@@ -231,8 +231,9 @@ typedef struct MQTTDeserializedInfo
  * @note The #MQTTGetCurrentTimeFunc_t callback function must be defined. If
  * there is no time implementation, it is the responsibility of the application
  * to provide a dummy function to always return 0, and provide 0 timeouts for
- * functions. This will ensure all time based functions will run for a single
```
abhidixi11, line 234: Can we add similar @note on top of these three functions as well ?

abhidixi11 (approved):

n9wxu (approved):

PR #96, opened 2020-10-30
Remove the memcpy function body in MQTT_Connect and MQTT_SerializeConnect proofs.

sarenameas:
When developing the proofs I had added the memcpy stub to the list of files to compile. At the time, I noticed this didn't really help the run time. I just now realized that I had forgotten to remove the memcpy function body so that the proof can actually use the stub!


yourslab (approved):

markrtuttle (approved):
I have built the goto binaries and confirmed that the stubbed out memcpy is the version of memcpy being used.  Good catch!

PR #113, opened 2020-11-23
Cast variable type based on format identifier

yourslab:
Only ever use base format specifiers and always cast to the standard C type

### DIFF
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -950,7 +950,7 @@ static MQTTStatus_t sendPublishAcks( MQTTContext_t * pContext,
             LogError( ( "Failed to send ACK packet: PacketType=%02x, "
                         "SentBytes=%d, "
                         "PacketSize=%lu.",
-                        packetTypeByte,
+                        ( int ) packetTypeByte,
                         ( int ) bytesSent,
                         MQTT_PUBLISH_ACK_PACKET_SIZE ) );
             status = MQTTSendFailed;
@@ -1198,7 +1198,7 @@ static MQTTStatus_t handleIncomingAck( MQTTContext_t * pContext,
 
     appCallback = pContext->appCallback;
 
-    LogDebug( ( "Received packet of type %02x.", pIncomingPacket->type ) );
+    LogDebug( ( "Received packet of type %02x.", ( int ) pIncomingPacket->type ) );
 
     switch( pIncomingPacket->type )
     {
@@ -1233,7 +1233,7 @@ static MQTTStatus_t handleIncomingAck( MQTTContext_t * pContext,
         default:
             /* Bad response from the server. */
             LogError( ( "Unexpected packet type from server: PacketType=%02x.",
-                        pIncomingPacket->type ) );
+                        ( int ) pIncomingPacket->type ) );
             status = MQTTBadResponse;
             break;
     }
@@ -1501,8 +1501,8 @@ static MQTTStatus_t receiveConnack( const MQTTContext_t * pContext,
         {
             LogError( ( "Incorrect packet type %X received while expecting"
                         " CONNACK(%X).",
-                        pIncomingPacket->type,
-                        MQTT_PACKET_TYPE_CONNACK ) );
+                        ( int ) pIncomingPacket->type,
+                        ( int ) MQTT_PACKET_TYPE_CONNACK ) );
             status = MQTTBadResponse;
         }
     }
@@ -2356,7 +2356,8 @@ MQTTStatus_t MQTT_GetSubAckStatusCodes( const MQTTPacketInfo_t * pSubackPacket,
     {
         LogError( ( "Invalid parameter: Input packet is not a SUBACK packet: "
                     "ExpectedType=%02x, InputType=%02x",
-                    MQTT_PACKET_TYPE_SUBACK, pSubackPacket->type ) );
+                    ( int ) MQTT_PACKET_TYPE_SUBACK,
+                    ( int ) pSubackPacket->type ) );
         status = MQTTBadParameter;
     }
     else if( pSubackPacket->pRemainingData == NULL )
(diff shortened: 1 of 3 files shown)


source/core_mqtt_serializer.c
```diff
@@ -1272,7 +1276,8 @@ static MQTTStatus_t deserializePublish( const MQTTPacketInfo_t * pIncomingPacket
         {
             *pPacketId = UINT16_DECODE( pPacketIdentifierHigh );
 
-            LogDebug( ( "Packet identifier %hu.", *pPacketId ) );
+            LogDebug( ( "Packet identifier %hu.",
+                        ( unsigned short int ) *pPacketId ) );
```
muneebahmed10, line 1280: Nit: you don't need the `int` here since it's already a `short`, and it would make it consistent with how `unsigned long` is used instead of `unsigned long int`

source/core_mqtt_serializer.c
```diff
@@ -2327,7 +2334,8 @@ MQTTStatus_t MQTT_DeserializeAck( const MQTTPacketInfo_t * pIncomingPacket,
 
             /* Any other packet type is invalid. */
             default:
-                LogError( ( "IotMqtt_DeserializeResponse() called with unknown packet type:(%02x).", pIncomingPacket->type ) );
+                LogError( ( "IotMqtt_DeserializeResponse() called with unknown packet type:(%02x).",
+                            pIncomingPacket->type ) );
```
muneebahmed10, line 2338: Shouldn't there be one here too?
```suggestion
                            ( unsigned int ) pIncomingPacket->type ) );
```

yourslab: Good point. I think it should be signed though.

source/core_mqtt_serializer.c
```diff
@@ -934,7 +934,7 @@ static MQTTStatus_t deserializeConnack( const MQTTPacketInfo_t * pConnack,
      * "Remaining length" of 2. */
     if( pConnack->remainingLength != MQTT_PACKET_CONNACK_REMAINING_LENGTH )
     {
-        LogError( ( "CONNACK does not have remaining length of %d.",
+        LogError( ( "CONNACK does not have remaining length of %u.",
                     MQTT_PACKET_CONNACK_REMAINING_LENGTH ) );
```
muneebahmed10, line 938: I believe this is defined is a `uint8_t`, so would it need to be cast as well?

source/core_mqtt_serializer.c
```diff
@@ -1319,7 +1325,7 @@ static MQTTStatus_t deserializeSimpleAck( const MQTTPacketInfo_t * pAck,
     /* Check that the "Remaining length" of the received ACK is 2. */
     if( pAck->remainingLength != MQTT_PACKET_SIMPLE_ACK_REMAINING_LENGTH )
     {
-        LogError( ( "ACK does not have remaining length of %d.",
+        LogError( ( "ACK does not have remaining length of %u.",
                     MQTT_PACKET_SIMPLE_ACK_REMAINING_LENGTH ) );
```
muneebahmed10, line 1329: Also a `uint8_t`

source/core_mqtt_serializer.c
yourslab: Yes, you are right. Did not notice the macro had a cast in it as well. Thank you for catching 👍🏽

muneebahmed10 (approved):

source/core_mqtt.c
```diff
@@ -605,7 +605,7 @@ static int32_t sendPacket( MQTTContext_t * pContext,
 
         if( bytesSent < 0 )
         {
-            LogError( ( "Transport send failed. Error code=%d.", bytesSent ) );
+            LogError( ( "Transport send failed. Error code=%d.", ( int ) bytesSent ) );
```
dan4thewin, line 608: Looks like bytesSent is an int32.  For the particular case which started this, the message was
`format '%d' expects argument of type 'int', but argument 4 has type 'int32_t' {aka 'long int'}`

I believe we want to use `%ld` and cast to long to handle the case where int is smaller that 32 bits.

yourslab: Good catch. I'll cast `int32_t` to `long int` and `int16_t` to `int`.

muneebahmed10: Why cast `int16_t` to `int`? A `short` is guaranteed to be at least 2 bytes.

Nit: `long int` and `long` are identical, I think we've been leaving off the `int` in other places where it's not necessary

yourslab: Oh yeah. I think I just wanted to err on the side of caution, but I didn't end up casting any `int16_t` types anyway.

source/core_mqtt.c
```diff
@@ -941,16 +941,15 @@ static MQTTStatus_t sendPublishAcks( MQTTContext_t * pContext,
 
             if( status != MQTTSuccess )
             {
-                LogError( ( "Failed to update state of publish %u.", packetId ) );
+                LogError( ( "Failed to update state of publish %u.",
+                            ( unsigned int ) packetId ) );
```
muneebahmed10, line 945: Since this is originally a `uint16_t`, should we use unsigned short?
```suggestion
                LogError( ( "Failed to update state of publish %hu.",
                            ( unsigned short ) packetId ) );
```

yourslab: I think the definition of `short` is really dependent on the compiler, so I wanted to err on the side of caution. Although, we use `short` in other places for 16-byte data types, so I have updated it.

muneebahmed10 (approved):

source/core_mqtt.c
muneebahmed10: An `unsigned short` is required to be able to hold values up to 65,535, according to [5.2.4.2](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf)

cobusve (approved):

PR #118, opened 2020-12-02
Use stdbool.h

muneebahmed10:
*Description*:
Removes the preprocessor check that relies on C99 macros and just includes `stdbool.h`. A `stdbool.readme` file, along with `stdint.readme`, is provided for compilers that do not provide their respective headers

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -147,7 +147,7 @@ jobs:
         run: |
           git-secrets --register-aws
           git-secrets --scan
-  custom-header:
+  custom-standard-c-headers:
     runs-on: ubuntu-latest
     steps:
       - name: Clone This Repo
@@ -159,7 +159,6 @@ jobs:
           cp source/include/stdint.readme override-include/stdint.h
           cmake -S test -B build/ \
           -G "Unix Makefiles" \
-          -DCMAKE_BUILD_TYPE=Debug \
           -DBUILD_CLONE_SUBMODULES=ON \
-          -DCMAKE_C_FLAGS='-Wall -Wextra -DNDEBUG -I../override-include'
+          -DCMAKE_C_FLAGS='-Wall -Wextra -I../override-include'
           make -C build/ coverity_analysis
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -147,3 +147,19 @@ jobs:
         run: |
           git-secrets --register-aws
           git-secrets --scan
+  custom-header:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone This Repo
+        uses: actions/checkout@v2
+      - name: Build
+        run: |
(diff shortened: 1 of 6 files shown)


source/include/stdbool.readme
```diff
@@ -0,0 +1,31 @@
...
+#ifndef __cplusplus
+
+/* _Bool was introduced in C99. */
+#define bool    int
+#define false   0
+#define true    1
+
+#endif
+
+#define __bool_true_false_are_defined   1
```
aggarw13, line 29: What is the significance of this macro?

muneebahmed10: It's part of the C standard to define this macro in stdbool.h along with the other three: https://pubs.opengroup.org/onlinepubs/009695399/basedefs/stdbool.h.html

source/include/stdbool.readme
```diff
@@ -0,0 +1,31 @@
+
+#ifndef _STDBOOL_H
```
aggarw13, line 2: Should there be a GitHub Action job to actually run the build check with the `stdbool.readme` and `stdint.readme` files to validate their definitions?

muneebahmed10: Added. Couldn't think of a good name for it though

.github/workflows/ci.yml
```diff
@@ -147,3 +147,19 @@ jobs:
         run: |
           git-secrets --register-aws
           git-secrets --scan
+  custom-header:
```
aggarw13, line 150:
```suggestion
  custom-standard-c-headers:
```

.github/workflows/ci.yml
```diff
@@ -147,3 +147,19 @@ jobs:
...
+      - name: Clone This Repo
+        uses: actions/checkout@v2
+      - name: Build
+        run: |
+          mkdir -p override-include
+          cp source/include/stdbool.readme override-include/stdbool.h
+          cp source/include/stdint.readme override-include/stdint.h
+          cmake -S test -B build/ \
+          -G "Unix Makefiles" \
+          -DCMAKE_BUILD_TYPE=Debug \
```
aggarw13, line 162: This is conflicting with the `-DNDEBUG` flag passed. Should that flag be passed?

muneebahmed10: I just copied the steps of the `unittest` job, we do this so that asserts don't show up in coverage. I could remove one or both and it shouldn't really affect anything since we're only testing if it builds without errors

aggarw13: Maybe the `unittest` job should mention the build type as `RELEASE` which (I think) would automatically take care of providing the `-DNDEBUG` flag to the compiler.
That is beyond the scope of this PR though, so will not block on it.

aggarw13 (dismissed):

aggarw13 (approved):

n9wxu (approved):

PR #121, opened 2020-12-03
Automate release with GitHub Action Job

aggarw13:
**Add a GitHub Action job to trigger a release process** which includes operations of: 
1. Creating and pushing the release tag to the repository 
2. Verifying the pushed tag (performing a diff with the commit ID which is tagged) 
3. Creating a ZIP for the release asset. 
4. Verifying the ZIP by performing a diff check and running unit tests on the unzipped files. 5. Creating a release on the repository for the tag, and uploading the ZIP as the release asset

The workflow can be manually triggered and takes the input values of **Commit ID** (to create a release for) and the **Version string** for tagging the release with

**Testing**
Here is an example run of the release job on my fork repository: 
https://github.com/aggarw13/coreMQTT/actions/runs/399119805. 
It pushed the test tag (`test-v2`) and created a release along with the release asset on the repository: https://github.com/aggarw13/coreMQTT/releases/tag/test-v2


.github/workflows/release-workflow.yml
```diff
@@ -0,0 +1,115 @@
...
+          ref: ${{ github.event.inputs.commit_id }}
+          path: coreMQTT
+          submodules: recursive
+      - name: Checkout disabled submodules
+        run: |
+          cd coreMQTT
+          git submodule update --init --checkout --recursive
+      - name: Create ZIP
+        run: |
+          zip -r coreMQTT-${{ github.event.inputs.version_number }}.zip coreMQTT -x "*.git"
```
leegeth, line 54: This will leave the .github folder in the zip. We were not doing that for the prior releases. I think it is better to avoid it.
The command can be modified to do that
`zip -r coreMQTT-${{ github.event.inputs.version_number }}.zip coreMQTT -x "*.git*"`

aggarw13, line 54:
```suggestion
          zip -r coreMQTT-${{ github.event.inputs.version_number }}.zip coreMQTT -x "*.git*"
```

leegeth (approved):

cobusve (approved):

PR #124, opened 2020-12-05
Fixes in internal logic of sending packets

aggarw13:
**This PR fixes 2 different issues, one for `sendPacket` and one for `sendPublish` internal functions,** both related to sending packets over the network.

#### Issue 1
For `sendPacket`, there is a possibility of encountering an infinite loop if the `Transport_Send_t` always returns zero bytes sent. The PR addresses this issue by adding a timeout logic on retrying zero send attempts. The timeout can be configured through the `MQTT_SEND_RETRY_TIMEOUT_MS` macro. 

#### Issue 2
For the `sendPublish` function (that calls the `sendPacket` function to send the complete packet over the network), there is a bug of treating partial or zero bytes return value of `sendPacket` as success. This PR updates the logic to treat partial or zero transmission of packets as failure in the call to `MQTT_Publish` API.

### DIFF
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -696,7 +696,7 @@ static int32_t recvExact( const MQTTContext_t * pContext,
     uint8_t * pIndex = NULL;
     size_t bytesRemaining = bytesToRecv;
     int32_t totalBytesRecvd = 0, bytesRecvd;
-    uint32_t lastDataRecvTimeMs = 0U, noDataRecvdTimeMs = 0U;
+    uint32_t lastDataRecvTimeMs = 0U, timeSinceLastDataWasRecvdMs = 0U;
     TransportRecv_t recvFunc = NULL;
     MQTTGetCurrentTimeFunc_t getTimeStampMs = NULL;
     bool receiveError = false;
@@ -751,10 +751,12 @@ static int32_t recvExact( const MQTTContext_t * pContext,
         else
         {
             /* No bytes were read from the network. */
-            noDataRecvdTimeMs = calculateElapsedTime( getTimeStampMs(), lastDataRecvTimeMs );
+            timeSinceLastDataWasRecvdMs =
+                calculateElapsedTime( getTimeStampMs(), lastDataRecvTimeMs );
         }
 
-        if( ( bytesRemaining > 0U ) && ( noDataRecvdTimeMs >= MQTT_RECV_POLLING_TIMEOUT_MS ) )
+        if( ( bytesRemaining > 0U ) &&
+            ( timeSinceLastDataWasRecvdMs >= MQTT_RECV_POLLING_TIMEOUT_MS ) )
         {
             LogError( ( "Time expired while receiving packet." ) );
             receiveError = true;
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -696,7 +696,7 @@ static int32_t recvExact( const MQTTContext_t * pContext,
     uint8_t * pIndex = NULL;
     size_t bytesRemaining = bytesToRecv;
     int32_t totalBytesRecvd = 0, bytesRecvd;
-    uint32_t lastDataRecvTimeMs = 0U, timeSinceLastRecvdMs = 0U;
+    uint32_t lastDataRecvTimeMs = 0U, timeSinceLastRecvMs = 0U;
     TransportRecv_t recvFunc = NULL;
     MQTTGetCurrentTimeFunc_t getTimeStampMs = NULL;
     bool receiveError = false;
@@ -751,11 +751,11 @@ static int32_t recvExact( const MQTTContext_t * pContext,
         else
         {
             /* No bytes were read from the network. */
-            timeSinceLastRecvdMs = calculateElapsedTime( getTimeStampMs(), lastDataRecvTimeMs );
+            timeSinceLastRecvMs = calculateElapsedTime( getTimeStampMs(), lastDataRecvTimeMs );
         }
 
         if( ( bytesRemaining > 0U ) &&
-            ( timeSinceLastRecvdMs >= MQTT_RECV_POLLING_TIMEOUT_MS ) )
+            ( timeSinceLastRecvMs >= MQTT_RECV_POLLING_TIMEOUT_MS ) )
         {
             LogError( ( "Time expired while receiving packet." ) );
             receiveError = true;
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -80,8 +80,8 @@ static MQTTPubAckType_t getAckFromPacketType( uint8_t packetType );
  * repeatedly to read bytes from the network until either:
  * 1. The requested number of bytes @a bytesToRecv are read.
  *                    OR
- * 2. There is a timeout of MQTT_RECV_POLLING_TIMEOUT_MS duration
- * between receiving bytes over the network.
+ * 2. No data is received from the network for MQTT_RECV_POLLING_TIMEOUT_MS duration.
+ * 
  *                    OR
  * 3. There is an error in reading from the network.
  *
@@ -696,7 +696,7 @@ static int32_t recvExact( const MQTTContext_t * pContext,
     uint8_t * pIndex = NULL;
     size_t bytesRemaining = bytesToRecv;
     int32_t totalBytesRecvd = 0, bytesRecvd;
-    uint32_t entryTimeMs = 0U, noDataRecvdTimeMs = 0U;
+    uint32_t lastDataRecvTimeMs = 0U, noDataRecvdTimeMs = 0U;
     TransportRecv_t recvFunc = NULL;
     MQTTGetCurrentTimeFunc_t getTimeStampMs = NULL;
     bool receiveError = false;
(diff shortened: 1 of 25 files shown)


test/unit-test/core_mqtt_utest.c
```diff
@@ -1949,17 +2011,21 @@ void test_MQTT_ProcessLoop_Timer_Overflow( void )
...
     setupTransportInterface( &transport );
+    setupNetworkBuffer( &networkBuffer );
 
     networkBuffer.size = 1000;
     incomingPacket.type = MQTT_PACKET_TYPE_PUBLISH;
     incomingPacket.remainingLength = MQTT_SAMPLE_REMAINING_LENGTH;
 
     globalEntryTime = UINT32_MAX - MQTT_OVERFLOW_OFFSET;
-    expectedFinalTime = MQTT_TIMER_CALLS_PER_ITERATION * numIterations - MQTT_OVERFLOW_OFFSET;
+    expectedFinalTime = globalEntryTime + ( numIterations * MQTT_TIMER_CALLS_PER_ITERATION ) + 1;
```
muneebahmed10, line 2028: Nit: Why this change? I liked the way it looked before, where we calculate the expected time independently. If we're testing if an overflow calculation works as expected, it looks a little weird to me that we're doing the exact same overflow calculation here (I know since they're unsigned it's not really an "overflow" and addition is well-defined, and that these two expressions will always be identical for uint32_t, but you know what I mean).

test/unit-test/core_mqtt_utest.c
```diff
@@ -1949,17 +2011,21 @@ void test_MQTT_ProcessLoop_Timer_Overflow( void )
     MQTTPublishState_t publishState = MQTTPubAckSend;
     MQTTPublishState_t ackState = MQTTPublishDone;
     uint8_t i = 0;
-    uint8_t numIterations = ( MQTT_TIMER_OVERFLOW_TIMEOUT_MS / MQTT_TIMER_CALLS_PER_ITERATION ) + 1;
+    uint8_t numIterations = ( ( MQTT_TIMER_OVERFLOW_TIMEOUT_MS % MQTT_TIMER_CALLS_PER_ITERATION ) == 0 ) ?
+                            ( MQTT_TIMER_OVERFLOW_TIMEOUT_MS / MQTT_TIMER_CALLS_PER_ITERATION ) :
+                            ( ( MQTT_TIMER_OVERFLOW_TIMEOUT_MS / MQTT_TIMER_CALLS_PER_ITERATION ) + 1 );
```
muneebahmed10, line 2016: Is this just ceil( a / b )?  I believe something like `( MQTT_TIMER_OVERFLOW_TIMEOUT_MS + MQTT_TIMER_CALLS_PER_ITERATION - 1 ) / MQTT_TIMER_CALLS_PER_ITERATION` would be equivalent and somewhat shorter, right? And a comment explaining that this is just a / b rounded up?

source/core_mqtt.c
```diff
@@ -1386,7 +1408,7 @@ static MQTTStatus_t sendPublish( MQTTContext_t * pContext,
                             pContext->networkBuffer.pBuffer,
                             headerSize );
 
-    if( bytesSent < 0 )
+    if( bytesSent < ( int32_t ) headerSize )
```
muneebahmed10, line 1411: Are you going to make the same change in the other places where `sendPacket` is called, or will that be a separate PR?

source/include/core_mqtt_config_defaults.h
```diff
@@ -129,6 +129,28 @@
...
+ *
+ * When sending an MQTT packet, the transport send function may be called multiple
+ * times until all of the required number of bytes are sent.
+ * This timeout represents the maximum duration that is allowed for no data
+ * transmission over the network through the transport send function.
+ *
+ * If the timeout expires, the #MQTT_ProcessLoop and #MQTT_ReceiveLoop functions
+ * return #MQTTSendFailed.
+ *
+ * <b>Possible values:</b> Any positive integer up to SIZE_MAX. Recommended to
```
muneebahmed10, line 145: Now that I think about it, why do we say up to SIZE_MAX if it will only be compared to a uint32_t? On our dev systems, isn't that 2^ 64 - 1? Shouldn't this (and others) be "any positive 32 bit integer"?

source/interface/transport_interface.h
```diff
@@ -183,7 +183,10 @@ typedef int32_t ( * TransportRecv_t )( NetworkContext_t * pNetworkContext,
  * @param[in] pBuffer Buffer containing the bytes to send over the network stack.
  * @param[in] bytesToSend Number of bytes to send over the network.
  *
- * @return The number of bytes sent or a negative error code.
+ * @return The number of bytes sent or a negative value to indicate error.
+ * If no data is transmitted over the network due to a full TX buffer and
```
muneebahmed10, line 187: Nit: We used `@note`here when making a similar comment for TransportRecv_t. Also I think there would have to be an empty line between the `@return` and `@note` blocks, judging from the other doxygen comments

test/unit-test/core_mqtt_utest.c
aggarw13: This is correcting the calculation of the expected time counter after the call to `MQTT_ProcessLoop`. The `+1` is for the call made to `getTime` before all iterations of the loop in the API.

source/core_mqtt.c
aggarw13: Separate PR to avoid scope screep in this PR.

source/include/core_mqtt_config_defaults.h
aggarw13: Sure, good point. Made the change

source/interface/transport_interface.h
aggarw13: Made the API doc for send and receive consistent.

test/unit-test/core_mqtt_utest.c
aggarw13: Good suggestion, have made the suggested changes.

test/unit-test/core_mqtt_utest.c
muneebahmed10: What's the correction? This value is the same as before isn't it? 
`(UINT32_MAX - MQTT_OVERFLOW_OFFSET) + (numIterations * MQTT_TIMER_CALLS_PER_ITERATION ) + 1` is the same as 
`((0) - MQTT_OVERFLOW_OFFSET) + (numIterations * MQTT_TIMER_CALLS_PER_ITERATION)` or
`MQTT_TIMER_CALLS_PER_ITERATION * numIterations - MQTT_OVERFLOW_OFFSET`

Also, basing `expectedFinalTime` off of `globalEntryTime` results in a calculation that always yields the final value, regardless if the counter rolls over back to 0 or not, which is what this test seems to be testing (e.g. If someone accidentally changed the value of `globalEntryTime` so there's no longer a rollover, this test would still pass).

Ultimately it doesn't matter, since the value is the same as it was before, but that's why this change looked weird to me

source/core_mqtt.c
```diff
@@ -628,20 +640,30 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+        }
+        else
+        {
+            /* No bytes were sent over the network. */
+            timeSinceLastSendMs = calculateElapsedTime( pContext->getTime(), lastSendTimeMs );
+
+            /* Check for timeout if we have been waiting to send any data over the network. */
+            if( timeSinceLastSendMs >= MQTT_SEND_RETRY_TIMEOUT_MS )
+            {
+                LogError( ( "Unable to send packet: Timed out in sending any data through transport send." ) );
```
muneebahmed10, line 655: It sounds like it uses "send" one too many times
```suggestion
                LogError( ( "Unable to send packet: Timed out in transport send." ) );
```

source/core_mqtt.c
```diff
@@ -755,7 +777,7 @@ static int32_t recvExact( const MQTTContext_t * pContext,
             /* Check for timeout if we have been waiting to receive any byte on the network. */
             if( timeSinceLastRecvMs >= MQTT_RECV_POLLING_TIMEOUT_MS )
             {
-                LogError( ( "Time expired while receiving packet." ) );
+                LogError( ( "Unable to receive packet: Timed out in receiving any data from transport recv." ) );
```
muneebahmed10, line 780: Same
```suggestion
                LogError( ( "Unable to receive packet: Timed out in transport recv." ) );
```

muneebahmed10 (approved):

muneebahmed10 (approved):

test/unit-test/core_mqtt_utest.c
aggarw13: >Also, basing expectedFinalTime off of globalEntryTime results in a calculation that always yields the final value, regardless if the counter rolls over back to 0 or not,

This is the main motivation from a resiliency and readability perspective.

gkwicker (approved):

test/unit-test/core_mqtt_utest.c
muneebahmed10: I was saying it's less resilient since before, accidentally setting `globalEntryTime` incorrectly could be caught due to the test failing. Now, due to how `expectedFinalTime` is calculated from `globalEntryTime`, the test will always pass regardless of the initial value of the entry time

PR #125, opened 2020-12-06
Enable debug logging in unit-test build and fix logging compilation error

yourslab:
- Update `core_mqtt_config.h` to specify logging level as `DEBUG`
- Fix `LogDebug` in `recvExact`

### DIFF
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -743,8 +743,10 @@ static int32_t recvExact( const MQTTContext_t * pContext,
             totalBytesRecvd += ( int32_t ) bytesRecvd;
             pIndex += bytesRecvd;
             LogDebug( ( "BytesReceived=%ld, BytesRemaining=%lu, "
+                        "TotalBytesReceived=%ld.",
                         ( long int ) bytesRecvd,
-                        ( unsigned long ) bytesRemaining ) );
+                        ( unsigned long ) bytesRemaining,
+                        ( long int ) totalBytesRecvd ) );
         }
         else
         {
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -742,7 +742,7 @@ static int32_t recvExact( const MQTTContext_t * pContext,
             bytesRemaining -= ( size_t ) bytesRecvd;
             totalBytesRecvd += ( int32_t ) bytesRecvd;
             pIndex += bytesRecvd;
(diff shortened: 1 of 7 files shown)


abhidixi11 (dismissed):
Please find out why logging was disabled for this target.

source/core_mqtt.c
```diff
@@ -742,9 +742,10 @@ static int32_t recvExact( const MQTTContext_t * pContext,
             bytesRemaining -= ( size_t ) bytesRecvd;
             totalBytesRecvd += ( int32_t ) bytesRecvd;
             pIndex += bytesRecvd;
-            LogDebug( ( "BytesReceived=%ld, BytesRemaining=%lu, "
+            LogDebug( ( "BytesReceived=%ld, BytesRemaining=%lu, TotalBytesReceived=%ld.",
                         ( long int ) bytesRecvd,
-                        ( unsigned long ) bytesRemaining ) );
+                        ( unsigned long ) bytesRemaining,
+                        ( long int ) totalBytesRecvd ) );
```
muneebahmed10, line 748: Wasn't this removed due to complexity, or is it fine now?

test/unit-test/core_mqtt_config.h
```diff
@@ -30,6 +30,32 @@
...
+
+#include "logging_levels.h"
+
+/* Logging configuration for the MQTT library. */
+#ifndef LIBRARY_LOG_NAME
+    #define LIBRARY_LOG_NAME    "MQTT"
+#endif
+
+#ifndef LIBRARY_LOG_LEVEL
+    #define LIBRARY_LOG_LEVEL    LOG_DEBUG
```
muneebahmed10, line 52: Would it make more sense to pass this from the command line, so that anyone running the unit tests does not have to change this manually?

source/core_mqtt.c
yourslab: It was but by keeping the string on the same line, the complexity is kept at 8. I'm working on using sed to remove the logs before checking complexity though. I don't think it is ideal to make logs less informative to satisfy complexity.

test/unit-test/core_mqtt_config.h
yourslab: I like that. I've made the change.

muneebahmed10 (approved):

cobusve (approved):

PR #127, opened 2020-12-07
Update APIs to detect partial sends as failures

aggarw13:
All library API functions call `sendPacket` to send an entire MQTT packet over the network. 
With the addition of a send timeout (with `MQTT_SEND_RETRY_TIMEOUT_MS` configuration) in PR #124, the APIs should detect partial packet (or zero bytes) transmission as send failures. 

This PR updates the `MQTT_Connect`, `MQTT_Disconnect`, `MQTT_Subscribe`, `MQTT_Unsubscibe` and `MQTT_Ping` API functions to return `MQTTSendFailed` for partial packet transmissions. 

**Note:** The `MQTT_Publish` function was updated for the same already in PR#124

### DIFF
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -696,7 +696,7 @@ static int32_t recvExact( const MQTTContext_t * pContext,
     uint8_t * pIndex = NULL;
     size_t bytesRemaining = bytesToRecv;
     int32_t totalBytesRecvd = 0, bytesRecvd;
-    uint32_t lastDataRecvTimeMs = 0U, noDataRecvdTimeMs = 0U;
+    uint32_t lastDataRecvTimeMs = 0U, timeSinceLastDataWasRecvdMs = 0U;
     TransportRecv_t recvFunc = NULL;
     MQTTGetCurrentTimeFunc_t getTimeStampMs = NULL;
     bool receiveError = false;
@@ -751,10 +751,12 @@ static int32_t recvExact( const MQTTContext_t * pContext,
         else
         {
             /* No bytes were read from the network. */
-            noDataRecvdTimeMs = calculateElapsedTime( getTimeStampMs(), lastDataRecvTimeMs );
+            timeSinceLastDataWasRecvdMs =
+                calculateElapsedTime( getTimeStampMs(), lastDataRecvTimeMs );
         }
 
-        if( ( bytesRemaining > 0U ) && ( noDataRecvdTimeMs >= MQTT_RECV_POLLING_TIMEOUT_MS ) )
+        if( ( bytesRemaining > 0U ) &&
+            ( timeSinceLastDataWasRecvdMs >= MQTT_RECV_POLLING_TIMEOUT_MS ) )
         {
             LogError( ( "Time expired while receiving packet." ) );
             receiveError = true;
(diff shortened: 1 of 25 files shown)


test/unit-test/core_mqtt_utest.c
```diff
@@ -2215,6 +2233,15 @@ void test_MQTT_Subscribe_error_paths( void )
...
     mqttStatus = MQTT_Subscribe( &context, &subscribeInfo, 1, MQTT_FIRST_VALID_PACKET_ID );
     TEST_ASSERT_EQUAL( MQTTSendFailed, mqttStatus );
+
+    /* Case when there is timeout in sending data through transport send. */
+    transport.send = transportSendNoBytes; /* Use the mock function that returns zero bytes sent. */
+    MQTT_GetSubscribePacketSize_ExpectAnyArgsAndReturn( MQTTSuccess );
+    MQTT_GetSubscribePacketSize_ReturnThruPtr_pPacketSize( &packetSize );
+    MQTT_GetSubscribePacketSize_ReturnThruPtr_pRemainingLength( &remainingLength );
+    MQTT_SerializeSubscribe_ExpectAnyArgsAndReturn( MQTTSuccess );
+    mqttStatus = MQTT_Subscribe( &context, &subscribeInfo, 1, MQTT_GetPacketId( &context ) );
```
muneebahmed10, line 2243: We probably shouldn't use an API that we're not testing
```suggestion
    mqttStatus = MQTT_Subscribe( &context, &subscribeInfo, 1, MQTT_FIRST_VALID_PACKET_ID );
```

aggarw13: Sure, have updated to use `MQTT_FIRST_VALID_PACKET_ID`

muneebahmed10 (approved):

source/core_mqtt.c
```diff
@@ -1796,7 +1796,7 @@ MQTTStatus_t MQTT_Connect( MQTTContext_t * pContext,
                                 pContext->networkBuffer.pBuffer,
                                 packetSize );
 
-        if( bytesSent < 0 )
+        if( bytesSent < ( int32_t ) packetSize )
```
gkwicker, line 1799: This test collapses both error conditions (`bytesSent < 0` and `bytesSent < packetSize`), so the error message on line 1801 is not as helpful as it should be.  If `bytesSent < 0` the actual error returned from the underlying transport is discarded – this can be very important when debugging so it should be logged.

It looks like these error messages had the same problem in the previous revision.  Let's improve them to make it easier to debug networking issues.

Same comment at 1890, 2023, 2085, & 2135.

gkwicker (changes requested):
See comment

source/core_mqtt.c
aggarw13: For a negative value error code returned by the `Transport_Send_t` call, the error code is logged in the `sendPacket` function here: https://github.com/FreeRTOS/coreMQTT/blob/main/source/core_mqtt.c#L625
Thus, as long as the logging is enabled for the library at `LOG_ERROR` level, the return code from the networking stack will be logged.

Also, if there was a timeout in sending the packet, that will also be logged separately by the function.

gkwicker: What if it's HTTP?

gkwicker (approved):

PR #129, opened 2020-12-09
Update Changelog for v1.1.0

muneebahmed10:
*Description*:
Changelog updates since v1.0.1 release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,8 @@
 
 ### Updates
  - [#118](https://github.com/FreeRTOS/coreMQTT/pull/118) Use the `stdbool.h` header file instead of using preprocessor checks for when `bool` is not defined. This also provides `stdbool.readme` and `stdint.readme` files in the case that a non-C99 compiler does not provide the respective header.
- - [#120](https://github.com/FreeRTOS/coreMQTT/pull/120) Introduce a `MQTT_RECV_POLLING_TIMEOUT_MS` macro to control the timeout before a multi-part transport read operation returns an error. Previously, it was controlled by a runtime timeout parameter, so an error might be returned when data could still be read. Now, reads will wait at least the macro timeout value before returning error. Conversely, reads will also wait at most the macro timeout value if no data is received in that duration, regardless of the timeout passed at runtime.
- - [#124](https://github.com/FreeRTOS/coreMQTT/pull/124), [#127](https://github.com/FreeRTOS/coreMQTT/pull/127), Introduce a `MQTT_SEND_RETRY_TIMEOUT_MS` macro to control the same timeout for transport sends.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
 # Changelog for coreMQTT Client Library
 
+## v1.0.2 (December 2020)
```
sarenameas, line 3: This needs to be 1.1.0
```suggestion
## v1.1.0 (December 2020)
```

abhidixi11 (dismissed):

CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
 # Changelog for coreMQTT Client Library
 
+## v1.1.0 (December 2020)
+
+### Updates
+ - [#118](https://github.com/FreeRTOS/coreMQTT/pull/118) Use the `stdbool.h` header file instead of using preprocessor checks for when `bool` is not defined. This also provides `stdbool.readme` and `stdint.readme` files in the case that a non-C99 compiler does not provide the respective header.
+ - [#120](https://github.com/FreeRTOS/coreMQTT/pull/120) Introduce a `MQTT_RECV_POLLING_TIMEOUT_MS` macro to control the timeout before a multi-part transport read operation returns an error. Previously, it was controlled by a runtime timeout parameter, so an error might be returned when data could still be read. Now, reads will wait at least the macro timeout value before returning error. Conversely, reads will also wait at most the macro timeout value if no data is received in that duration, regardless of the timeout passed at runtime.
```
aggarw13, line 7: > Conversely, reads will also wait at most the macro timeout value if no data is received in that duration, regardless of the timeout passed at runtime.
This is not necessarily true. Even if a zero value for the timeout is used, the maximum blocking time would be dependent on the socket timeout of the underlying transport read implementation

```suggestion
 - [#120](https://github.com/FreeRTOS/coreMQTT/pull/120) Introduce a `MQTT_RECV_POLLING_TIMEOUT_MS` configuration macro to control the timeout for retrying zero byte network read operations before returning an error. Previously, all network read attempts were timed out with a runtime parameter, that could result in a premature timeout and MQTT disconnection even when data could still be read. Now, reads will wait at least the configured macro timeout value before returning error.
```

CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
...
+ - [#124](https://github.com/FreeRTOS/coreMQTT/pull/124), [#127](https://github.com/FreeRTOS/coreMQTT/pull/127), Introduce a `MQTT_SEND_RETRY_TIMEOUT_MS` macro to control the same timeout for transport sends.
```
aggarw13, line 8:
```suggestion
 - [#124](https://github.com/FreeRTOS/coreMQTT/pull/124), [#127](https://github.com/FreeRTOS/coreMQTT/pull/127), Introduce a `MQTT_SEND_RETRY_TIMEOUT_MS` configuration macro to control the timeout for retrying zero byte transport send operations.
```

CHANGELOG.md
muneebahmed10: It could be reworded but it is true. The timeout is the maximum duration during which no data may be received. Previously, if the process loop were called with say, 500 ms, then we could spend up to those 500 ms receiving data. If a partial receive errored, then it was possible to just increase the process loop timeout for the next time, which could even be done at runtime. Now, we may still spend up to 500 ms or even more, but only if the maximum interval between successful reads doesn't exceed the timeout. if a partial receive errors, we need to increase the macro value and recompile. We saw this happen during testing for the MQTT agent.

abhidixi11 (approved):

n9wxu (approved):

PR #132, opened 2020-12-09
Add doc that keep-alive is unsupported with dummy timer

aggarw13:
Update API doc to mention that the **keep-alive** mechanism is not supported by `MQTT_ProcessLoop` API when a dummy timer function that always returns zero is supplied to the library.

### DIFF
--- a/source/include/core_mqtt.h
+++ b/source/include/core_mqtt.h
@@ -609,12 +609,12 @@ MQTTStatus_t MQTT_Disconnect( MQTTContext_t * pContext );
  * run, unless an error occurs.
  *
  * @note Calling this function blocks the calling context for a time period that
- * depends on the passed @p timeoutMs, the configuration macro, #MQTT_RECV_POLLING_TIMEOUT_MS,
- * and the underlying transport interface implementation timeouts, unless an error
- * occurs.
- *    Blocking Time = Max( timeoutMs parameter,
- *                         MQTT_RECV_POLLING_TIMEOUT_MS,
- *                         Transport interface send/recv implementation timeout )
+ * depends on the passed @p timeoutMs, the configuration macros, #MQTT_RECV_POLLING_TIMEOUT_MS
+ * and #MQTT_SEND_RETRY_TIMEOUT_MS, and the underlying transport interface implementation
+ * timeouts, unless an error occurs. The blocking period also depends on the execution time of the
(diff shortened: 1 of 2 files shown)


source/include/core_mqtt_config_defaults.h
```diff
@@ -100,6 +100,12 @@
  * If a ping response is not received before this timeout, then
  * #MQTT_ProcessLoop will return #MQTTKeepAliveTimeout.
  *
+ * @note If a dummy implementation of the #MQTTGetCurrentTimeFunc_t timer function,
```
abhidixi11, line 103: 1. Let's just say that if a dummy implementation of the timer function is used. I don't we need to go into details of 'that always returns zero'. Let's keep it simple.
2.  processLoop is not supported without a 'real' time function. In that case receiveLoop should be used. ( receive loop is exactly processLopp mius keep alive.

source/include/core_mqtt.h
```diff
@@ -598,21 +598,23 @@ MQTTStatus_t MQTT_Disconnect( MQTTContext_t * pContext );
  * alive.
  *
  * @note Passing a timeout value of 0 will run the loop for a single iteration.
- * If a dummy #MQTTGetCurrentTimeFunc_t was passed to #MQTT_Init, then the timeout
```
abhidixi11, line 601: We should just say that it is not supported with dummy timer function and reference doxygen for timer function ( to explain what we mean by dummy)

source/include/core_mqtt_config_defaults.h
aggarw13: Have incorporated point 1. The second point is not appropriate for this config that is only utilized by ProcessLoop.

source/include/core_mqtt.h
aggarw13: The presence of `#MQTTGetCurrentTimeFunc_t` adds the reference to the documentation of the timer function

aggarw13: >We should just say that it is not supported with dummy timer function 
Only the keep-alive part of the function is not supported with a dummy timer function. The rest of the functionality of receiving acks and incoming PUBLISH is supported

abhidixi11 (dismissed):

abhidixi11 (approved):

dan4thewin (approved):

PR #136, opened 2020-12-09
Update API documentation of transport send/recv functions

aggarw13:
Update API of the transport interface functions for the following: 
* Transport recv() should NOT block when requested to read a single byte. It MAY block for the underlying socket timeout when requested to read n > 1 bytes.
* Transport send() should NOT return an error when the send fails because of a full TX buffer of the underlying network stack, so that the calling library can retry the send operation.

### DIFF
--- a/source/interface/transport_interface.h
+++ b/source/interface/transport_interface.h
@@ -1,5 +1,5 @@
 /*
- * coreMQTT v1.0.1
+ * coreMQTT v1.1.0
  * Copyright (C) 2020 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy of
--- a/source/interface/transport_interface.h
+++ b/source/interface/transport_interface.h
@@ -105,7 +105,7 @@
  *        // to avoid blocking for socket timeout.
  *        if( TLSRecvCount( pNetworkContext->tlsContext ) == 0 )
  *        {
- *            callTlsRecvFunct = false;
+ *            callTlsRecvFunc = false;
  *        }
  *     }
  *
@@ -199,7 +199,7 @@ typedef struct NetworkContext NetworkContext_t;
  *
  * @note If no data is available on the network to read and no error
  * has occurred, zero MUST be the return value. A zero return value
- * SHOULD represent that the read operation can be retried by caling
+ * SHOULD represent that the read operation can be retried by calling
  * the API function. Zero MUST NOT be returned if a network disconnection
  * has occurred.
  */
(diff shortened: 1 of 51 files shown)


source/interface/transport_interface.h
```diff
@@ -96,16 +96,31 @@
...
+ *
+ *     // For a single byte read request, check if data is available on the network.
+ *     if( bytesToRecv == 1 )
  *     {
- *         // Handle socket error.
+ *        // If no data is available on the network, do not call TLSRecv
+ *        // to avoid blocking for socket timeout.
+ *        if( TLSRecvCount( pNetworkContext->tlsContext ) == 0 )
+ *        {
+ *            callTlsRecvFunct = false;
```
sarenameas:
```suggestion
 *            callTlsRecvFunc = false;
```

sarenameas: Also why can you just shorten this into a an OR condition like https://github.com/FreeRTOS/FreeRTOS/pull/452/files#diff-988c68b1458ff5f8870190348da08357275ca5b5286dfd6776a236166829b113R131

aggarw13: For easier readability for the reader

sarenameas (dismissed):

dan4thewin (approved):

source/interface/transport_interface.h
```diff
@@ -96,16 +96,31 @@
...
  *
+ *     if( callTlsRecvFunc == true )
+ *     {
+ *        bytesReceived = TLSRecv( pNetworkContext->tlsContext,
+ *                                 pBuffer,
+ *                                 bytesToRecv,
+ *                                 MY_SOCKET_TIMEOUT );
+ *        if( bytesReceived < 0 )
+ *        {
+ *            // Handle socket error.
```
RichardBarry, line 120: Here if bytesReceived contains an error code that just means the TCP read timed (or would block) then that is not an error so bytesReceived needs to be translated into whatever the calling function will interpret as "try again" - which for our code is zero, but if this function is calling the TCP read function and passing a value back to the TLS stack then zero may also be interpreted as an error.

source/interface/transport_interface.h
```diff
@@ -132,7 +147,14 @@
  *                          pBuffer,
  *                          bytesToSend,
  *                          MY_SOCKET_TIMEOUT );
- *     if( bytesSent < 0 )
+ *
+ *      // If underlying TCP buffer is full, set the return value to zero
+ *      // so that caller can retry operation.
+ *     if( bytesSent == TCP_BUFFER_FULL )
+ *     {
+ *          bytesSent = 0;
```
RichardBarry, line 155: Similar comment as per the receive - when the error is not really an error, rather just a full buffer, the error has to be translated into whatever the calling function interprets as meaning 'try again' until you time out.

aggarw13: Yes, the zero return value represents that the operation can be re-tried.

source/interface/transport_interface.h
aggarw13: Have updated the pseudo code block to add a check for a retriable error.

RichardBarry (approved):

sarenameas (approved):

PR #148, opened 2021-01-11
Update doxyfile for doxygen 1.9.2

muneebahmed10:
*Description*:
`doxygen` 1.9.2 deprecates the following tags, resulting in warnings:
* 'OUTPUT_TEXT_DIRECTION'
* 'COLS_IN_ALPHA_INDEX'
* 'LATEX_SOURCE_CODE'
* 'RTF_SOURCE_CODE'
* 'DOCBOOK_PROGRAMLISTING'

The first two commits update the file manually to allow it to work with both 1.8.20 and 1.9.2 without warnings. The last commit updates the file with `doxygen -u`, but it will only work with 1.9.2. I don't think there's any value in keeping compatibility with 1.8.20, so I think it's better to just use `doxygen -u`.


docs/doxygen/config.doxyfile
```diff
@@ -1594,11 +1607,29 @@ FORMULA_TRANSPARENT    = YES
...
+# With MATHJAX_VERSION it is possible to specify the MathJax version to be used.
+# Note that the different versions of MathJax have different requirements with
+# regards to the different settings, so it is possible that also other MathJax
+# settings have to be changed when switching between the different MathJax
+# versions.
+# Possible values are: MathJax_2 and MathJax_3.
+# The default value is: MathJax_2.
+# This tag requires that the tag USE_MATHJAX is set to YES.
+
+MATHJAX_VERSION        = MathJax_2
```
aggarw13, line 1619: As `USE_MATHJAX` is set to `NO`, should this be set to an empty value to avoid giving misrepresentation of MathJax being used in the doxygen generation?

docs/doxygen/config.doxyfile
```diff
@@ -533,6 +532,13 @@ EXTRACT_LOCAL_METHODS  = NO
 
 EXTRACT_ANON_NSPACES   = NO
 
+# If this flag is set to YES, the name of an unnamed parameter in a declaration
+# will be determined by the corresponding definition. By default unnamed
+# parameters remain unnamed in the output.
+# The default value is: YES.
+
+RESOLVE_UNNAMED_PARAMS = YES
```
aggarw13, line 540: Should this be set to `NO` instead to discourage using unnamed parameters in function declarations?

docs/doxygen/config.doxyfile
```diff
@@ -2366,10 +2378,32 @@ UML_LOOK               = NO
...
+# tag is set to YES, doxygen will add type and arguments for attributes and
+# methods in the UML graphs. If the DOT_UML_DETAILS tag is set to NONE, doxygen
+# will not generate fields with class member information in the UML graphs. The
+# class diagrams will look similar to the default class diagrams but using UML
+# notation for the relationships.
+# Possible values are: NO, YES and NONE.
+# The default value is: NO.
+# This tag requires that the tag UML_LOOK is set to YES.
+
+DOT_UML_DETAILS        = NO
```
aggarw13, line 2396: As the description for this config seems relevant to OOP languages, should this be set to `NONE` for our libraries?

muneebahmed10:
The changes to these files were done by `doxygen -u`. If we make additional changes here we should propagate them to all the other repos

docs/doxygen/config.doxyfile
muneebahmed10: MISRA already warns us about unnamed parameters

archigup (approved):

paulbartell (approved):

PR #159, opened 2021-04-01
Add check for 0 packet ID for subacks

muneebahmed10:
*Description*:

`core_mqtt_serializer.c` checks for invalid packet IDs for all packets except for SUBACKs. Since packets are received from the network, we should check to make sure they are not malformed.
This adds a zero packet ID check for subacks, adds an appropriate unit test, and updates the change log.

The license file had execute permissions for some reason. This removes them.


source/core_mqtt_serializer.c
```diff
@@ -1164,8 +1164,15 @@ static MQTTStatus_t deserializeSuback( const MQTTPacketInfo_t * pSuback,
         LogDebug( ( "Packet identifier %hu.",
                     ( unsigned short ) *pPacketIdentifier ) );
 
-        status = readSubackStatus( remainingLength - sizeof( uint16_t ),
-                                   pVariableHeader + sizeof( uint16_t ) );
+        if( *pPacketIdentifier == 0U )
```
aggarw13, line 1167: Should the `MQTT_PACKET_ID_INVALID` macro be moved the serializer API?https://github.com/FreeRTOS/coreMQTT/blob/main/source/include/core_mqtt.h#L55-L61
```suggestion
        if( *pPacketIdentifier == MQTT_PACKET_ID_INVALID )
```

muneebahmed10: Oh wow, I forgot about this. Possibly, but doing that is a larger change than what this PR was focused on.

aggarw13 (approved):

paulbartell (approved):

PR #163, opened 2021-06-29
Fix keep alive checking interval

muneebahmed10:
*Description*:
As pointed out in https://forums.freertos.org/t/keepalivetimeout-code-in-coremqtt/12957/3, when a ping request is sent from `MQTT_ProcessLoop()`, it is only checked for a response after the keep alive interval has elapsed again. Since the broker would have disconnected the client by that time anyway if the PINGREQ was not received, the response ought to be checked earlier, as dictated by `MQTT_PINGRESP_TIMEOUT_MS`.

Two solutions for this issue are:
1. Update `handleKeepAlive` to check based on `pingReqSendTimeMs` and not `lastPacketTimeMs`.
2. Update `MQTT_Ping()` to only update `pingReqSendTimeMs` and not `lastPacketTimeMs`.

The latter solution would mean that sending other packets, e.g. a PUBLISH, would reset the interval before the PINGRESP is checked, while in the former it would not. ~However, the latter solution results in a smaller binary when optimizing with `-O1`~. This PR implements the former solution, and leaves discussion for the tradeoffs for this review

### DIFF
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -995,6 +995,7 @@ static MQTTStatus_t handleKeepAlive( MQTTContext_t * pContext )
 {
     MQTTStatus_t status = MQTTSuccess;
     uint32_t now = 0U, keepAliveMs = 0U;
+    uint32_t pingrespTimeoutMs = MQTT_PINGRESP_TIMEOUT_MS;
 
     assert( pContext != NULL );
     assert( pContext->getTime != NULL );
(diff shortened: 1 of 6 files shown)


source/include/core_mqtt_config_defaults.h
```diff
@@ -106,11 +106,11 @@
...
  *
  * <b>Possible values:</b> Any positive integer up to SIZE_MAX. <br>
- * <b>Default value:</b> `500`
+ * <b>Default value:</b> `5000`
  */
 #ifndef MQTT_PINGRESP_TIMEOUT_MS
-    /* Wait 0.5 seconds by default for a ping response. */
-    #define MQTT_PINGRESP_TIMEOUT_MS    ( 500U )
+    /* Wait 5 seconds by default for a ping response. */
+    #define MQTT_PINGRESP_TIMEOUT_MS    ( 5000U )
```
muneebahmed10, line 113: I thought 5 seconds was a more reasonable default for this, for slower network conditions. From just testing locally with QEMU, it takes > 1 second for a CONNACK to be received, and sometimes with device advisor, the CONNACK takes 5-10 seconds to arrive. Since this bugfix will make this macro actually be used, we might be getting more bug reports from customers if we don't update this value now too.

source/core_mqtt.c
```diff
@@ -995,29 +995,38 @@ static MQTTStatus_t handleKeepAlive( MQTTContext_t * pContext )
...
 
     assert( pContext != NULL );
     assert( pContext->getTime != NULL );
 
     now = pContext->getTime();
     keepAliveMs = 1000U * ( uint32_t ) pContext->keepAliveIntervalSec;
 
+    /* It is not useful for the ping response timeout to be greater than half
+     * the keep alive, as the server will have disconnected us by that point. */
+    if( pingrespTimeoutMs > ( keepAliveMs / 2 ) )
```
aggarw13, line 1008: Should this be prevented as a compile time build failure instead of changing it at runtime and the customer not being aware of the use of a different response timeout than their configured value?

muneebahmed10: I don't see how this could be done at compile time, since the keep alive information is set in the connect info that's passed to `MQTT_Connect()`. The best that could be done is a warning log

aggarw13 (approved):

n9wxu (approved):

PR #168, opened 2021-07-22
Update version for release

muneebahmed10:
*Description*:
Updates version number for release, and adds guards for C++ linkage. The `/* *INDENT-OFF* */` is necessary for uncrustify. Also updates the documentation links to point to repositories/latest instead of specific releases, and adds a memory estimates table.

The documentation link for the CSDK points to the repository's README as it aggregates links for all releases. Since the FreeRTOS/FreeRTOS repository does not, the link here points to the latest hosted documentation on freertos.org

Also updates `test/CMakeLists.txt` so that tests can be built with newer C standards if specified on the command line


README.md
```diff
@@ -122,6 +122,13 @@ Please refer to the demos of the MQTT client library in the following locations
 
 ## Generating documentation
 
+Note: For pre-generated documentation, please see the documentation linked in the above repositories:
```
aggarw13, line 125: It may be better to have the table under a different section than "Generating Documentation" as Abhijit had mentioned that the main concern is that it makes it misleading for customers about them having to generate documentation. 

If the table is present in a separate section, like "Existing Documentation", toward the beginning of the README, it would be easier for customers to notice

yourslab (approved):

README.md
```diff
@@ -122,6 +122,13 @@ Please refer to the demos of the MQTT client library in the following locations
...
+
+| Location |
+| :-: |
+| [AWS IoT Device SDK for Embedded C](https://github.com/aws/aws-iot-device-sdk-embedded-C#releases-and-documentation) |
+| [FreeRTOS.org](https://freertos.org/Documentation/api-ref/coreMQTT/docs/doxygen/output/html/index.html) |
```
aggarw13, line 130: If the released versions of C-SDK and FreeRTOS.org don't match, it would be confusing to customers about the reason of their disparity. It would be important to add a note saying that it is possible for the C-SDK and FreeRTOS releases to contain different release versions of coreMQTT

README.md
muneebahmed10: How about this?

# Documentation

## Existing Documentation

## Generating Documentation

aggarw13: Yes, that's fine. These can be moved before the Unit Tests section though to make it more prominent for customers to find

yourslab (dismissed):

README.md
```diff
@@ -120,7 +120,20 @@ Please refer to the demos of the MQTT client library in the following locations
 | FreeRTOS | [FreeRTOS AWS Reference Integrations](https://github.com/aws/amazon-freertos/tree/master/demos/coreMQTT) | Based on Secure Sockets Abstraction |
 
 
-## Generating documentation
+## Documentation
+
+### Existing Documentation
+
+For pre-generated documentation, please see the documentation linked in the above repositories:
```
yourslab, line 127:
```suggestion
For pre-generated documentation, please see the documentation linked in the repositories below:
```

muneebahmed10: Fixed

yourslab (approved):

yourslab (approved):

yourslab (approved):

n9wxu (dismissed):

yourslab (dismissed):

yourslab (approved):

n9wxu (approved):

PR #169, opened 2021-07-23
testing expensive proofs

ronakfof:


ronakfof:
I tested running the proofs for coreMQTT on Codebuild instance of general1.large instead of general1.2xlarge.
Runtime increased by more than double from the usual ~10minutes to roughly **~22minutes.**
This means the status checks on the PR would take roughly twice as long than usual.
Assuming an average of ~80 executions per month, the cost is expected to decrease by ~$109 by switching from general1.2xlarge to general1.large.

test/cbmc/proofs/MQTT_Connect/Makefile
```diff
@@ -92,4 +92,6 @@ PROJECT_SOURCES += $(SRCDIR)/source/core_mqtt.c
 PROJECT_SOURCES += $(SRCDIR)/source/core_mqtt_serializer.c
 PROJECT_SOURCES += $(SRCDIR)/source/core_mqtt_state.c
 
+EXPENSIVE = true
```
yourslab, line 95: Just curious as to what exactly this does? How do you know what proofs should be marked as `EXPENSIVE`?

yourslab (approved):

test/cbmc/proofs/MQTT_Connect/Makefile
ronakfof: Setting EXPENSIVE to those particular proofs allows those proofs to be run separately (each one on their own) whereas the rest run in the usual parallelized fashion.
MQTT_ProcessLoop was EXPENSIVE as seen [**HERE**](https://d2s5j4r0gkya4b.cloudfront.net/final/4c742b5e-6010-4ac3-843f-6133e0158be6/pipelines/MQTT_ProcessLoop/index.html#job-c820f83e-14ad-4663-9f88-6ead71deaaab).
I did the same for two other proofs which were taking longer as seen on the runtime graph on [**THIS**](https://d2s5j4r0gkya4b.cloudfront.net/final/4c742b5e-6010-4ac3-843f-6133e0158be6/index.html) page.

yourslab: Thanks @ronakfof for the thorough explanation. I'll also let each repository owner check the memory usage graphs so that they can appropriately mark certain proofs as `EXPENSIVE`.

paulbartell (approved):

PR #178, opened 2021-10-18
Removes all instances of mallocCanFail from CBMC proofs

feliperodri:
Signed-off-by: Felipe R. Monteiro <felisous@amazon.com>

We don't need to use `mallocCanFail` anymore. CBMC non-deterministically fails `malloc` when invoked with `--malloc-may-fail` and `--malloc-fail-null` flags (see `Makefile.common`).

### DIFF
--- a/test/cbmc/sources/mqtt_cbmc_state.c
+++ b/test/cbmc/sources/mqtt_cbmc_state.c
@@ -69,6 +69,11 @@ bool isValidMqttPacketInfo( const MQTTPacketInfo_t * pPacketInfo )
 {
     bool isValid = true;
 
+    if( pPacketInfo != NULL )
+    {
+        isValid = isValid && pPacketInfo->remainingLength < REMAINING_LENGTH_MAX;
+    }
+
     return isValid;
 }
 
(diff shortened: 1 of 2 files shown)


source/core_mqtt_serializer.c
```diff
@@ -519,7 +519,7 @@ static uint8_t * encodeString( uint8_t * pDestination,
     pBuffer++;
 
     /* Copy the string into pBuffer. */
-    if( pSourceBuffer != NULL )
+    if( ( pSourceBuffer != NULL ) && ( sourceLength != 0 ) )
```
markrtuttle, line 522: Why does a pull request removing mallocCanFail from proofs make a modification to the project code?

test/cbmc/sources/mqtt_cbmc_state.c
```diff
@@ -50,25 +50,19 @@
     #define REMAINING_LENGTH_MAX    CBMC_MAX_OBJECT_SIZE
 #endif
 
-void * mallocCanFail( size_t size )
-{
-    __CPROVER_assert( size < CBMC_MAX_OBJECT_SIZE, "mallocCanFail size is too big" );
```
markrtuttle, line 55: Can you confirm that this project is using all of the malloc flags, and that these malloc flags replicate this assertion failure being deleted here?

source/core_mqtt_serializer.c
feliperodri: This change is unnecessary, I reverted it.

PR #190, opened 2022-04-06
Reduce severity of logs

NathanJPhillips:
Messages about disconnection/connection can perhaps be at info level as they are rarely printed but publish/acknowledge messages can occur very frequently during normal operation and probably don't need to be output at such a high level. There are four log messages in particular that are printed at info level that occur very frequently in normal operation, clogging up log files and preventing other messages standing out. This PR changes those messages to debug level.

### DIFF
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -1051,7 +1051,7 @@ static MQTTStatus_t handleIncomingPublish( MQTTContext_t * pContext,
     assert( pContext->appCallback != NULL );
 
     status = MQTT_DeserializePublish( pIncomingPacket, &packetIdentifier, &publishInfo );
-    LogInfo( ( "De-serialized incoming PUBLISH packet: DeserializerResult=%s.",
+    LogDebug( ( "De-serialized incoming PUBLISH packet: DeserializerResult=%s.",
                MQTT_Status_strerror( status ) ) );
 
     if( status == MQTTSuccess )
@@ -1064,7 +1064,7 @@ static MQTTStatus_t handleIncomingPublish( MQTTContext_t * pContext,
 
         if( status == MQTTSuccess )
         {
-            LogInfo( ( "State record updated. New state=%s.",
+            LogDebug( ( "State record updated. New state=%s.",
                        MQTT_State_strerror( publishRecordState ) ) );
         }
 
@@ -1168,7 +1168,7 @@ static MQTTStatus_t handlePublishAcks( MQTTContext_t * pContext,
 
     ackType = getAckFromPacketType( pIncomingPacket->type );
     status = MQTT_DeserializeAck( pIncomingPacket, &packetIdentifier, NULL );
-    LogInfo( ( "Ack packet deserialized with result: %s.",
+    LogDebug( ( "Ack packet deserialized with result: %s.",
                MQTT_Status_strerror( status ) ) );
 
     if( status == MQTTSuccess )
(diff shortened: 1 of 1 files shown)


archigup:
Could you describe the your specific use case or issue you are facing?
I see that there are only these logs and connect/disconnect in the info log level, and they seem to fit there.

If the concern is that there is too much output, a better solution is likely to set the log level for coreMQTT to LOG_WARN or LOG_ERROR

NathanJPhillips:
The issue is the semantic meaning of Info vs Debug. It's often a subjective choice, but if you're a user of the library rather than a library developer, then messages telling you that the library is doing the thing that your function call told it to do are probably more properly considered debug level messages rather than info level messages.

NathanJPhillips:
I may need some help with fixing the formatting CI failure. Can someone point out the relevant error message in the CI log?

source/core_mqtt.c
```diff
@@ -1064,7 +1064,7 @@ static MQTTStatus_t handleIncomingPublish( MQTTContext_t * pContext,
 
         if( status == MQTTSuccess )
         {
-            LogInfo( ( "State record updated. New state=%s.",
+            LogDebug( ( "State record updated. New state=%s.",
                        MQTT_State_strerror( publishRecordState ) ) );
```
archigup, line 1068:
```suggestion
                        MQTT_State_strerror( publishRecordState ) ) );
```

source/core_mqtt.c
```diff
@@ -1168,7 +1168,7 @@ static MQTTStatus_t handlePublishAcks( MQTTContext_t * pContext,
 
     ackType = getAckFromPacketType( pIncomingPacket->type );
     status = MQTT_DeserializeAck( pIncomingPacket, &packetIdentifier, NULL );
-    LogInfo( ( "Ack packet deserialized with result: %s.",
+    LogDebug( ( "Ack packet deserialized with result: %s.",
                MQTT_Status_strerror( status ) ) );
```
archigup, line 1172:
```suggestion
                MQTT_Status_strerror( status ) ) );
```

source/core_mqtt.c
```diff
@@ -1181,7 +1181,7 @@ static MQTTStatus_t handlePublishAcks( MQTTContext_t * pContext,
 
         if( status == MQTTSuccess )
         {
-            LogInfo( ( "State record updated. New state=%s.",
+            LogDebug( ( "State record updated. New state=%s.",
```
archigup, line 1184:
```suggestion
             LogDebug( ( "State record updated. New state=%s.",
```

source/core_mqtt.c
```diff
@@ -1181,7 +1181,7 @@ static MQTTStatus_t handlePublishAcks( MQTTContext_t * pContext,
...
                        MQTT_State_strerror( publishRecordState ) ) );
```
archigup, line 1185:
```suggestion
                        MQTT_State_strerror( publishRecordState ) ) );
```

archigup:
I added suggestions for the formatting issues

archigup:
In our experience with helping customers debug and configure the MQTT library, we have found that logging incoming packets has been valuable for debugging. Seeing when the packets arrive in the logs helps determine if the library is properly configured or if you are encountering networking issues. Developers using the library should be concerned with error and warn level logs which signal issues to the developers attention; enabling info logging is for when one is interested in the state changes within the system and info about what the system is doing as we do here. Debug logs are generally far more detailed info mostly useful for debugging particular MQTT issues.

The logging level should be easily configurable on a per-library basis, and by moving to LOG_WARN for coreMQTT these logs are already disabled, so moving them to debug would reduce the flexibility in choosing logging level.

NathanJPhillips:
@archigup thank you for taking the time to clearly explain your rationale. Much appreciated.

PR #191, opened 2022-05-31
generate PINGREQ packets on idle input or output.

adam-scislowicz:
Extend the PINGREQ logic to send a PINGREQ not only when the client has not generated any outgoing message for the Keep Alive period (ref: [3.1.2.10 Keep Alive](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc385349238) (http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc385349238)). But to also generate a PINGREQ if we have not received any incoming message for the Keep Alive period, then in the case that we are continuously retransmitting a PUBLISH qos1 but receiving no ACK, we will still generate a PINGREQ after the specified Keep Alive delay. The spec explicitly allows a PINGREQ to be sent at any time and so this change in behavior maintains compliance.

Additional configuration options have been added (via the header file). This allows for the specification of different timeouts on idle transmit (TX) vs receive (RX) activity.

The smaller of the runtime configuration keepAliveTimeout and compile time transmit (TX) time timeout are used as the transmit timeout.

### DIFF
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -1024,8 +1024,11 @@ static MQTTStatus_t handleKeepAlive( MQTTContext_t * pContext )
         }
         else
         {
-            if( calculateElapsedTime( now, pContext->lastPacketTime ) > keepAliveMs )
+            if(( calculateElapsedTime( now, pContext->lastPacketTime ) > keepAliveMs ) ||
+              ( calculateElapsedTime( now, pContext->lastPacketReceivedTime) > keepAliveMs ))
             {
+                pContext->lastPacketReceivedTime = now; /* this to avoid sending back-to-back MQTT pings. Note that MQTT_Ping will
+                                                         * internally update the lastPacketTime as it calls sendPacket() */
                 status = MQTT_Ping( pContext );
             }
         }
@@ -1340,6 +1343,7 @@ static MQTTStatus_t receiveSingleIteration( MQTTContext_t * pContext,
     if( status == MQTTSuccess )
     {
         incomingPacket.pRemainingData = pContext->networkBuffer.pBuffer;
+        pContext->lastPacketReceivedTime = pContext->getTime();
 
         /* PUBLISH packets allow flags in the lower four bits. For other
          * packet types, they are reserved. */
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -1002,37 +1002,33 @@ static MQTTStatus_t sendPublishAcks( MQTTContext_t * pContext,
 static MQTTStatus_t handleKeepAlive( MQTTContext_t * pContext )
 {
     MQTTStatus_t status = MQTTSuccess;
-    uint32_t now = 0U, keepAliveMs = 0U, packetRxTimeoutMs = 0U;
+    uint32_t now = 0U, packetTxTimeoutMs = 0U, packetRxTimeoutMs = 0U;
 
     assert( pContext != NULL );
     assert( pContext->getTime != NULL );
 
     now = pContext->getTime();
-    keepAliveMs = 1000U * ( uint32_t ) pContext->keepAliveIntervalSec;
-    packetRxTimeoutMs = 1000U * ( uint32_t ) pContext->packetRxTimeoutSec;
+    packetTxTimeoutMs = MIN(PACKET_TX_TIMEOUT_MS, 1000U * ( uint32_t ) pContext->keepAliveIntervalSec);
 
     /* If keep alive interval is 0, it is disabled. */
-    if( keepAliveMs != 0U )
+    if( pContext->waitingForPingResp == true )
     {
-        if( pContext->waitingForPingResp == true )
+        /* Has time expired? */
+        if( calculateElapsedTime( now, pContext->pingReqSendTimeMs ) >
+            MQTT_PINGRESP_TIMEOUT_MS )
         {
-            /* Has time expired? */
-            if( calculateElapsedTime( now, pContext->pingReqSendTimeMs ) >
-                MQTT_PINGRESP_TIMEOUT_MS )
-            {
-                status = MQTTKeepAliveTimeout;
-            }
+            status = MQTTKeepAliveTimeout;
         }
-        else
+    }
+    else
+    {
+        if((packetTxTimeoutMs != 0U) && (calculateElapsedTime( now, pContext->lastPacketTxTime ) >= packetTxTimeoutMs) )
         {
-            if( calculateElapsedTime( now, pContext->lastPacketTxTime ) >= keepAliveMs )
-            {
-                status = MQTT_Ping( pContext );
-            }
-            else if( calculateElapsedTime( now, pContext->lastPacketRxTime) >= packetRxTimeoutMs )
-            {
-                status = MQTT_Ping( pContext );
-            }
+            status = MQTT_Ping( pContext );
+        }
+        else if((PACKET_RX_TIMEOUT_MS != 0U) && ( calculateElapsedTime( now, pContext->lastPacketRxTime) >= PACKET_RX_TIMEOUT_MS) )
+        {
+            status = MQTT_Ping( pContext );
         }
     }
 
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -1008,11 +1008,14 @@ static MQTTStatus_t handleKeepAlive( MQTTContext_t * pContext )
     assert( pContext->getTime != NULL );
 
     now = pContext->getTime();
-    packetTxTimeoutMs = 1000U * ( uint32_t ) pContext->keepAliveIntervalSec;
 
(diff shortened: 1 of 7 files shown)


source/core_mqtt.c
```diff
@@ -1024,8 +1024,11 @@ static MQTTStatus_t handleKeepAlive( MQTTContext_t * pContext )
         }
         else
         {
-            if( calculateElapsedTime( now, pContext->lastPacketTime ) > keepAliveMs )
+            if(( calculateElapsedTime( now, pContext->lastPacketTime ) > keepAliveMs ) ||
```
paulbartell, line 1027:
```suggestion
            if(( calculateElapsedTime( now, pContext->lastPacketTime ) >= keepAliveMs ) ||
```

source/core_mqtt.c
```diff
@@ -1024,8 +1024,11 @@ static MQTTStatus_t handleKeepAlive( MQTTContext_t * pContext )
...
+              ( calculateElapsedTime( now, pContext->lastPacketReceivedTime) > keepAliveMs ))
```
paulbartell, line 1028:
```suggestion
              ( calculateElapsedTime( now, pContext->lastPacketReceivedTime) >= keepAliveMs ))
```

source/core_mqtt.c
```diff
@@ -1024,8 +1024,11 @@ static MQTTStatus_t handleKeepAlive( MQTTContext_t * pContext )
...
             {
+                pContext->lastPacketReceivedTime = now; /* this to avoid sending back-to-back MQTT pings. Note that MQTT_Ping will
```
paulbartell, line 1030: Not sure this is necessary because pContext->waitingForPingResp will always be true once a ping has been sent.

source/include/core_mqtt.h
```diff
@@ -212,6 +212,11 @@ typedef struct MQTTContext
      */
     uint32_t lastPacketTime;
 
+    /**
+     * @brief Timestamp of the last packet received by the library.
+     */
+    uint32_t lastPacketReceivedTime;
```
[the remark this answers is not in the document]
adam-scislowicz: I can do this, but it will require further refactoring I think.

source/core_mqtt.c
```diff
@@ -1024,8 +1024,11 @@ static MQTTStatus_t handleKeepAlive( MQTTContext_t * pContext )
         }
         else
         {
-            if( calculateElapsedTime( now, pContext->lastPacketTime ) > keepAliveMs )
+            if(( calculateElapsedTime( now, pContext->lastPacketTxTime ) > keepAliveMs ) ||
```
paulbartell, line 1027:
```suggestion
            if( ( pContext->waitingForPingResp == false ) && 
                 ( ( calculateElapsedTime( now, pContext->lastPacketTxTime ) > keepAliveMs ) ||
```

source/core_mqtt.c
```diff
@@ -1024,8 +1024,11 @@ static MQTTStatus_t handleKeepAlive( MQTTContext_t * pContext )
...
+              ( calculateElapsedTime( now, pContext->lastPacketRxTime) > keepAliveMs ))
             {
+                pContext->lastPacketRxTime = now; /* this to avoid sending back-to-back MQTT pings. Note that MQTT_Ping will
```
paulbartell, line 1030:
```suggestion
```

source/core_mqtt.c
```diff
@@ -1024,8 +1024,11 @@ static MQTTStatus_t handleKeepAlive( MQTTContext_t * pContext )
...
+                                                   * internally update the lastPacketTxTime as it calls sendPacket() */
```
paulbartell, line 1031:
```suggestion
```

source/core_mqtt.c
```diff
@@ -1024,7 +1025,11 @@ static MQTTStatus_t handleKeepAlive( MQTTContext_t * pContext )
         }
         else
         {
-            if( calculateElapsedTime( now, pContext->lastPacketTime ) > keepAliveMs )
+            if( calculateElapsedTime( now, pContext->lastPacketTxTime ) >= keepAliveMs )
```
ravibhagavandas, line 1028: Why should we check for both RX and TX timestamp expires ? IMO we should only check for RX timestamp expiry here as that would cover both cases.

source/core_mqtt.c
```diff
@@ -1002,13 +1002,14 @@ static MQTTStatus_t sendPublishAcks( MQTTContext_t * pContext,
...
     MQTTStatus_t status = MQTTSuccess;
-    uint32_t now = 0U, keepAliveMs = 0U;
+    uint32_t now = 0U, keepAliveMs = 0U, packetRxTimeoutMs = 0U;
 
     assert( pContext != NULL );
     assert( pContext->getTime != NULL );
 
     now = pContext->getTime();
     keepAliveMs = 1000U * ( uint32_t ) pContext->keepAliveIntervalSec;
+    packetRxTimeoutMs = 1000U * ( uint32_t ) pContext->packetRxTimeoutSec;
```
ravibhagavandas, line 1012: Can we use the same parameter as before for determining when to send a keep alive?

source/core_mqtt.c
```diff
@@ -1002,32 +1002,38 @@ static MQTTStatus_t sendPublishAcks( MQTTContext_t * pContext,
 static MQTTStatus_t handleKeepAlive( MQTTContext_t * pContext )
 {
     MQTTStatus_t status = MQTTSuccess;
-    uint32_t now = 0U, keepAliveMs = 0U;
+    uint32_t now = 0U, packetTxTimeoutMs = 0U, packetRxTimeoutMs = 0U;
```
paulbartell, line 1005: packetRxTimeoutMs is unused.
```suggestion
    uint32_t now = 0U, packetTxTimeoutMs = 0U;
```

source/core_mqtt.c
```diff
@@ -1002,32 +1002,38 @@ static MQTTStatus_t sendPublishAcks( MQTTContext_t * pContext,
...
 
     assert( pContext != NULL );
     assert( pContext->getTime != NULL );
 
     now = pContext->getTime();
-    keepAliveMs = 1000U * ( uint32_t ) pContext->keepAliveIntervalSec;
+    packetTxTimeoutMs = 1000U * ( uint32_t ) pContext->keepAliveIntervalSec;
+
+    if( PACKET_TX_TIMEOUT_MS < packetTxTimeoutMs )
```
paulbartell, line 1013: This might be a bit more clear.
```suggestion
    if( PACKET_TX_TIMEOUT_MS < ( 1000U * ( uint32_t ) pContext->keepAliveIntervalSec ) )
```

source/core_mqtt.c
```diff
@@ -1002,32 +1002,38 @@ static MQTTStatus_t sendPublishAcks( MQTTContext_t * pContext,
...
+    {
+        packetTxTimeoutMs = PACKET_TX_TIMEOUT_MS;
+    }
```
paulbartell, line 1017:
```suggestion
    else
    {
        packetTxTimeoutMs = ( 1000U * ( uint32_t ) pContext->keepAliveIntervalSec );
    }
```

paulbartell (approved):

source/core_mqtt.c
```diff
@@ -1002,32 +1002,39 @@ static MQTTStatus_t sendPublishAcks( MQTTContext_t * pContext,
...
-            if( calculateElapsedTime( now, pContext->lastPacketTime ) > keepAliveMs )
-            {
-                status = MQTT_Ping( pContext );
-            }
+            status = MQTT_Ping( pContext );
+        }
+        else if( ( PACKET_RX_TIMEOUT_MS != 0U ) && ( calculateElapsedTime( now, pContext->lastPacketRxTime ) >= PACKET_RX_TIMEOUT_MS ) )
+        {
+            status = MQTT_Ping( pContext );
         }
```
paulbartell, line 1038: For MISRA, each if/else if must have an else even if empty.

adam-scislowicz: got it, full logical coverage. I will adjust

paulbartell (approved):

source/include/core_mqtt_config_defaults.h
```diff
@@ -119,10 +119,36 @@
...
+/**
+ * @brief Maximum number of milliseconds of RX inactivity to wait
+ * before initiating a PINGREQ
+ *
+ * <b>Possible values:</b> Any positive integer up to SIZE_MAX. <br>
+ * <b>Default value:</b> '30000'
+ *
+ */
+#ifndef PACKET_RX_TIMEOUT_MS
+    #define PACKET_RX_TIMEOUT_MS    ( 25000U )
```
ravibhagavandas, line 149: Why is PACKET_RX_TIMEOUT_MS default value less than that of TX value ? Also default value comment does not match code.

paulbartell (approved):

PR #198, opened 2022-08-10
Add State to MQTT_ProcessLoop so that it can be called in a non-blocking manner

AniruddhaKanhere:


source/core_mqtt.c
```diff
@@ -851,6 +852,74 @@ static MQTTStatus_t discardPacket( const MQTTContext_t * pContext,
...
+    remainingLength = mqttPacketSize - pContext->index;
+
+    while( ( totalBytesReceived < remainingLength ) && ( receiveError == false ) )
+    {
+        if( ( remainingLength - totalBytesReceived ) < bytesToReceive )
+        {
+            bytesToReceive = remainingLength - totalBytesReceived;
+        }
+
+        bytesReceived = recvExact( pContext, bytesToReceive );
```
paulbartell, line 888: I still don't feel very comfortable with calling recv in a loop here.

paulbartell (approved):

source/core_mqtt.c
AniruddhaKanhere: True, but as we discussed, calling the `recv` function just once means that we'd need to maintain more state as we cannot receive a packet bigger than the buffer in the MQTT library.

So, this is a tradeoff which we can visit later. Does that sound reasonable?

Also, I'll be fixing the failed tests right now.

paulbartell (approved):

archigup (approved):

PR #200, opened 2022-09-02
Add hooks to the source code

AniruddhaKanhere:


archigup:
MQTT_GetPacketId also may need a mutex

archigup (dismissed):

archigup (dismissed):

archigup (approved):

source/core_mqtt.c
```diff
@@ -1269,22 +1285,31 @@ static MQTTStatus_t sendPublishAcks( MQTTContext_t * pContext,
...
                                     localBuffer.pBuffer,
                                     MQTT_PUBLISH_ACK_PACKET_SIZE );
+
+            MQTT_POST_SEND_HOOK( pContext );
         }
 
         if( bytesSent == ( int32_t ) MQTT_PUBLISH_ACK_PACKET_SIZE )
         {
             pContext->controlPacketSent = true;
+
```
paulbartell, line 1302: Would it be desirable to keep hold of the mutex between the sendPacket and MQTT_UpdateStateAck calls?

paulbartell:
Is a lock needed during calls to recv?

paulbartell:
Where is the user meant to store their mutex / lock context(s)?

Is it necessary to lock socket operations separately from state update operations?

source/core_mqtt.c
AniruddhaKanhere: I think not.
They are decoupled altogether. Also, the state will be protected by its mutex.

AniruddhaKanhere:
> Is a lock needed during calls to recv?

I think that receive should be done through one thread only. So, it doesn't seem like it would need to be locked.
There is one exception -> the call to MQTT_Connect will be receiving data by itself and not use processloop. But again, it doesn't make sense to do recv before connecting to the broker.

AniruddhaKanhere:
> Where is the user meant to store their mutex / lock context(s)?
> 
> Is it necessary to lock socket operations separately from state update operations?

The users are free to store the context wherever they like. To aid them to link the mutex with the context, the MQTT context is provided to the user through the macro's parameter.

I think since the send and state update ops are different and distinct, it is imperative that we separate them.

paulbartell (approved):

PR #203, opened 2022-09-06
Update MQTT logging so that log levels of the library do not leak

AniruddhaKanhere:
This PR defines the MQTT library log levels. They are defined as such that they do not leak into other libraries/applications which include core_mqtt.h

It is done in this way so that the API doesn't have to change.

Background: The core_mqtt.h header declares the `MQTTContext_t` structure which has configurable number of unacknowledged packets. This configuration comes from `core_mqtt_config.h`. This config file, however, also defines the logging levels. Thus, any library/application which includes `core_mqtt.h` (thereby including `core_mqtt_config.h`) gets the definitions of the logging macros. This is the leak that this PR aims to plug.

There were a few options that were considered:
- Asking the user for the memory for the unACKed packet list while calling MQTT_InIt. However, that might confuse the users and complicate the API
- Moving the logging macros to a private header which is included by the source files. This approach was discarded since this means that any user wanting to modify the logging level will need to modify with a private header file of the code base. This did not seem right at the time.

### DIFF
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -29,6 +29,12 @@
 
 #include "core_mqtt.h"
 #include "core_mqtt_state.h"
+
+#define LIBRARY_LOG_NAME "MQTT"
+#ifndef LIBRARY_LOG_LEVEL
+    #define LIBRARY_LOG_LEVEL MQTT_LOG_LEVEL
+#endif
+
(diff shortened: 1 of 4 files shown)


paulbartell (approved):

AniruddhaKanhere:
Closing this PR as this will not have the intended effect.

PR #211, opened 2022-09-12
Remove unnessesary ternarys

archigup:


source/core_mqtt.c
```diff
@@ -629,8 +628,7 @@ static bool matchWildcards( const char * pTopicName,
 
         /* If the topic name contains a child level but the topic filter ends at
          * the current level, then there does not exist a match. */
-        if( ( nextLevelExistsInTopicName == true ) &&
-            ( nextLevelExistsinTopicFilter == false ) )
+        if( nextLevelExistsInTopicName && !nextLevelExistsinTopicFilter )
```
AniruddhaKanhere, line 631: I would prefer the original to the modified version. The original seems more clear to me.

source/core_mqtt.c
```diff
@@ -641,7 +639,7 @@ static bool matchWildcards( const char * pTopicName,
          * can be performed in the next level.
          * Note: The name index already points to the level separator in the topic
          * name. */
-        else if( nextLevelExistsInTopicName == true )
+        else if( nextLevelExistsInTopicName )
```
AniruddhaKanhere, line 642: Here too. I would prefer the original.

source/core_mqtt.c
```diff
@@ -659,7 +657,7 @@ static bool matchWildcards( const char * pTopicName,
      * last character in a topic filter. */
     else if( ( pTopicFilter[ *pFilterIndex ] == '#' ) &&
              ( *pFilterIndex == ( topicFilterLength - 1U ) ) &&
-             ( locationIsValidForWildcard == true ) )
+             locationIsValidForWildcard )
```
AniruddhaKanhere, line 660: Here too. I would prefer the original.

AniruddhaKanhere (changes requested):
There are many places where we use a pattern: `if( xVariable == true )`. I prefer those over just using `if( xVariable )`.
The first one seems clearer to me. It might also help us catch bugs. Think of an extreme case when the bool variable was changed to a pointer. The new version would check for that pointer being NULL and compiler won't complain - in the older version, compiler would complain about it forcing us to fix it.

AniruddhaKanhere (approved):

alfred2g (approved):

PR #221, opened 2022-09-22
Remove restriction on LWT payload being zero

AniruddhaKanhere:


source/core_mqtt.c
```diff
@@ -1814,20 +1814,27 @@ static size_t addEncodedStringToVector( uint8_t serailizedLength[ 2 ],
...
+    packetLength = seralizedLengthFieldSize;
 
-    /* Then add the pointer to the string itself. */
-    pLocalIterator[ 1 ].iov_base = string;
-    pLocalIterator[ 1 ].iov_len = length;
-
-    packetLength = length + seralizedLengthFieldSize;
+    /* Sometimes the string can be NULL that is, of 0 length. In that case,
+     * only the length field should be encoded in the vector. */
+    if( ( string != NULL ) && ( length != 0U ) )
```
archigup, line 1830: What if string is null and length is non-zero? Should we just assert that string is non-null when length is nonzero?

AniruddhaKanhere: Oh true. Good catch

AniruddhaKanhere: Done!
That was a good catch.

paulbartell (approved):

Skptak (approved):

PR #223, opened 2022-09-29
Modify a check to make sure that keep alive is sent even when data is in the buffer

AniruddhaKanhere:
The coreMQTT library is responsible for sending the keep alive messages to the broker when there is no transaction of data between the broker and the client.

The check ( `if( status == MQTTNoDataAvailable )` ) for that condition is not correct anymore (since PR https://github.com/FreeRTOS/coreMQTT/pull/218) because this condition is only set when there is no data in the socket AND none in the receive buffer. Thus, if there is any data in the receive buffer, it would prevent the library from sending a ping.

But for the matter of keep alive, the server does not care whether there is data in the buffer or not. It only cares when was the last TCP transaction with the client which can be depicted by the data in the socket.

This PR updates the `if` condition such that the sending of keep alive is only based on the reception of data or lack thereof from the socket. The change is as follows:
```
-   if( status == MQTTNoDataAvailable ) // This status value is set only when there is no data in the buffer and not in the socket.
+   if( recvBytes == 0 ) // This value shows that no more data is available in the socket.
```

### DIFF
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -1695,6 +1695,11 @@ static MQTTStatus_t receiveSingleIteration( MQTTContext_t * pContext,
                 /* Reset the status. */
                 status = statusCopy;
             }
+            else
+            {
+                LogError( ( "Handling of keep alive failed. Status=%s",
+                            MQTT_Status_strerror( status ) ) );
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


archigup (approved):

paulbartell (approved):
Looks good to me.

archigup (approved):

paulbartell (approved):

PR #224, opened 2022-10-04
Fixes to timeout of sendMessageVector and refactor of sendBuffer for consistency

jasonpcarroll:
This PR fixes sendMessageVector improperly timing out and refactors both sendMessageVector and sendBuffer to be consistent with one another in their logic.

### DIFF
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -762,7 +762,7 @@ static int32_t sendMessageVector( MQTTContext_t * pContext,
      * while loop. */
     lastSendTimeMs = pContext->getTime();
 
-    while( ( bytesSentOrError < ( int32_t ) bytesToSend ) && ( bytesSentOrError >= 0 ) )
+    while( bytesSentOrError < ( int32_t ) bytesToSend )
     {
         if( pContext->transportInterface.writev != NULL )
         {
@@ -795,6 +795,7 @@ static int32_t sendMessageVector( MQTTContext_t * pContext,
         else if( sendResult < 0 )
         {
             bytesSentOrError = sendResult;
+            break;
         }
         else
         {
@@ -860,7 +861,7 @@ static int32_t sendBuffer( MQTTContext_t * pContext,
      * while loop. */
     lastSendTimeMs = pContext->getTime();
 
-    while( ( bytesSentOrError < ( int32_t ) bytesToSend ) && ( bytesSentOrError >= 0 ) )
+    while( bytesSentOrError < ( int32_t ) bytesToSend )
     {
         sendResult = pContext->transportInterface.send( pContext->transportInterface.pNetworkContext,
                                                         pIndex,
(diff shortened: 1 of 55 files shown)


source/core_mqtt.c
```diff
@@ -770,49 +772,65 @@ static int32_t sendMessageVector( MQTTContext_t * pContext,
...
-        /* Check for timeout. */
-        if( pContext->getTime() > timeoutTime )
-        {
-            break;
-        }
+    if( ( size_t ) bytesSentOrError == bytesToSend )
+    {
+        pContext->lastPacketTxTime = lastSendTimeMs;
+        LogDebug( ( "sendMessageVector: Successfully sent packet at time %lu.",
+                    ( unsigned long ) lastSendTimeMs ) );
```
AniruddhaKanhere, line 833: Should this be done even when we are able to send even a partial packet?

source/core_mqtt.c
```diff
@@ -733,35 +733,37 @@ static int32_t sendMessageVector( MQTTContext_t * pContext,
...
 {
-    uint32_t timeoutTime;
-    uint32_t bytesToSend = 0U;
-    int32_t bytesSentOrError = 0;
+    int32_t sendResult;
+    uint32_t lastSendTimeMs;
+    uint32_t timeSinceLastSendMs;
     TransportOutVector_t * pIoVectIterator;
     size_t vectorsToBeSent = ioVecCount;
+    size_t bytesToSend = 0;
```
AniruddhaKanhere, line 741: There should a `U` at the end here. That is, it should be:
`size_t bytesToSend = 0U;`

AniruddhaKanhere (changes requested):

source/core_mqtt.c
```diff
@@ -770,49 +772,65 @@ static int32_t sendMessageVector( MQTTContext_t * pContext,
...
+
+            LogDebug( ( "sendMessageVector: Bytes Sent=%ld, Bytes Remaining=%lu",
+                        ( long int ) sendResult,
+                        ( unsigned long ) ( bytesToSend - ( size_t ) bytesSentOrError ) ) );
         }
-        else
+        else if( sendResult < 0 )
         {
             bytesSentOrError = sendResult;
+        }
```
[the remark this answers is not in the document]
Skptak: Added in the empty else.

source/core_mqtt.c
Skptak: Added the ```U``` to this.

AniruddhaKanhere (changes requested):

source/include/core_mqtt_config_defaults.h
```diff
@@ -2,8 +2,6 @@
  * coreMQTT v2.0.0
  * Copyright (C) 2022 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
  *
- * SPDX-License-Identifier: MIT
- *
```
AniruddhaKanhere, line 6: Can you add this back please?

AniruddhaKanhere (changes requested):
Thank you for this PR @jasonpcarroll, can you make just one small change?

AniruddhaKanhere (approved):

Skptak (approved):

PR #227, opened 2022-10-13
Core mqtt migration guide update

jasonpcarroll:
This PR adds the migration guide for coreMQTT >=2.0.0.


Skptak (approved):

MigrationGuide.md
```diff
@@ -0,0 +1,224 @@
+## coreMQTT version >=v2.0.0 Migration Guide
+
+With coreMQTT versions >=v2.0.0, there are some breaking changes that need to be addressed when upgrading.
+
+### Breaking Changes
+
+The `MQTT_ProcessLoop` function no longer uses a timeout as this led to unavoidable busy-waiting. Thus, the signature of `MQTT_ProcessLoop` changed from `MQTTStatus_t MQTT_ProcessLoop( MQTTContext_t * pContext, uint32_t timeoutMs )`  to  `MQTTStatus_t MQTT_ProcessLoop( MQTTContext_t * pContext )`. Additionally, `MQTT_ProcessLoop` can now return `MQTTNeedMoreBytes`. A return of `MQTTNeedMoreBytes` means that `MQTT_ProcessLoop` received only a part of an MQTT packet and will need to be called again (probably after a bit of delay) in order to finish receiving the MQTT packet. Thus, to migrate, simply remove the timeout from the `MQTT_ProcessLoop` call and additionally check for if `MQTTNeedMoreBytes` has been returned when checking the status of `MQTT_ProcessLoop`. For example:
```
AniruddhaKanhere, line 7: Since these changes are not numbered (or added to bullets). They look like a big monolithic blob of text. I think we should try to separate them somehow.

Ref: take a look at the rendered markdown file here: https://github.com/FreeRTOS/coreMQTT/blob/coreMQTTMigration/MigrationGuide.md

MigrationGuide.md
```diff
@@ -0,0 +1,224 @@
...
+
+     if( status != MQTTSuccess && status != MQTTNeedMoreBytes )
+     {
+         // Determine the error. It's possible we might need to disconnect
+         // the underlying transport connection.
+     }
+     else
+     {
+         // Other application functions.
+     }
```
AniruddhaKanhere, line 51:
```suggestion
     /* else if only required if the terminating else is empty and the application doesn't want a scenario akin to busy waiting. */
     else if( status == MQTTNeedMoreBytes )
     {
         /* Only a partial MQTT packet is received. Call MQTT_ProcessLoop again; ideally after a small delay. */
     }
     else
     {
         // Other application functions.
     }
```

MigrationGuide.md
```diff
@@ -0,0 +1,224 @@
...
```
AniruddhaKanhere, line 99:
```suggestion
     /* else if only required if the terminating else is empty and the application doesn't want a scenario akin to busy waiting. */
     else if( status == MQTTNeedMoreBytes )
     {
         /* Only a partial MQTT packet is received. Call MQTT_ReceiveLoop again; ideally after a small delay. */
     }
     else
     {
         // Other application functions.
     }
```

MigrationGuide.md
```diff
@@ -0,0 +1,224 @@
...
+}
+```
+
+The `TransportInterface_t` structure now has a new member `writev` that can be defined to allow sending multiple MQTT message vectors as a single packet. This can improve performance, but is not necessary to implement. However, when programs initialize a `TransportInterface_t` structure, they **MUST** either set `writev` to a working implementation or set it `NULL`. Not doing this will lead to undefined behavior as the coreMQTT library checks if `writev` is `NULL` to determine if it should be used. For example:
```
AniruddhaKanhere, line 103:
```suggestion
The `TransportInterface_t` structure now has a new member `writev`. It uses scatter-gather approach to send multiple MQTT packet components as a single packet to reduce overhead and improve performance. However, it is COMPLETELY OPTIONAL to implement. To that end, when application(s) initialize a `TransportInterface_t` structure, they **MUST** either set `writev` to a working implementation or set it `NULL`. Not doing this will lead to undefined behavior as the coreMQTT library checks if `writev` is `NULL` to determine if it should be used. For example:
```

MigrationGuide.md
```diff
@@ -0,0 +1,224 @@
...
+}
+
+if( status == MQTTSuccess )
+{
+     // Do something with mqttContext. The transport and fixedBuffer structs were
+     // copied into the context, so the original structs do not need to stay in scope.
+}
+```
+
+For coreMQTT version >=v2.1.0, the `MQTT_SEND_RETRY_TIMEOUT_MS` macro configuration has been deprecated. It has been replaced with `MQTT_SEND_TIMEOUT_MS`. `MQTT_SEND_RETRY_TIMEOUT_MS` was formerly used to timeout sending an MQTT packet after the transport `send` function sent 0 bytes for too long. This timeout would reset each time the transport `send` function sent any bytes, leading to the actual timeout scaling with the number of bytes being sent. In other words, the maximum time for sending a packet was variable. In order to remedy this, the `MQTT_SEND_RETRY_TIMEOUT_MS` macro was replaced with `MQTT_SEND_TIMEOUT_MS`. `MQTT_SEND_TIMEOUT_MS` now sets the maximum duration that coreMQTT will attempt to send an MQTT packet, leading to more consistent timeout behavior.
```
AniruddhaKanhere, line 220:
```suggestion
For coreMQTT version >=v2.1.0, the `MQTT_SEND_RETRY_TIMEOUT_MS` macro configuration has been deprecated. It has been replaced with `MQTT_SEND_TIMEOUT_MS`. `MQTT_SEND_RETRY_TIMEOUT_MS` was formerly used to timeout sending an MQTT packet after the transport `send` function failed to transmit any bytes for too long. This timeout would reset each time the transport `send` function sent any bytes, leading to the actual timeout scaling with the number of bytes being sent. In other words, the maximum time for sending a packet was variable. In order to remedy this, the `MQTT_SEND_RETRY_TIMEOUT_MS` macro was replaced with `MQTT_SEND_TIMEOUT_MS`. `MQTT_SEND_TIMEOUT_MS` now sets the maximum duration that coreMQTT will attempt to send an MQTT packet, leading to more consistent timeout behavior across various APIs and packet lengths.
```

AniruddhaKanhere (changes requested):
Majorly, the formatting seems off when viewed in a rendered format.
There is no clear demarcation of where a description ends and where the next one begins.

MigrationGuide.md
jasonpcarroll: added bullets

AniruddhaKanhere (approved):

Skptak (approved):

PR #235, opened 2022-12-02
Fix to update the timestamp of the last packet received

onkwon:
It doesn't look like `pContext->lastPacketRxTime` getting updated anywhere. And once `PACKET_RX_TIMEOUT_MS` reached [the condition](https://github.com/FreeRTOS/coreMQTT/blob/19d198c8cb03c00307608b3008b62d433b6f4353/source/core_mqtt.c#L1358) will always be true.


jasonpcarroll:
Hi @onkwon,

Sorry for seeing this so late! Thank you for bringing this to our attention. I will take a look and notify other team members of this issue.

Best,

Jason Carroll

onkwon:
Please note that this would cause customer bill increase as ping-pong every `MQTT_ProcessLoop()` after `PACKET_RX_TIMEOUT_MS`.

archigup:
Hey, this PR needs a change in docs/doxygen/include/size_table.md (see [here](https://github.com/FreeRTOS/coreMQTT/actions/runs/3599283599/jobs/6064300062)). If you'd prefer, I can add the change to this PR for you.

Otherwise, PR looks good to me

AniruddhaKanhere (approved):
Thanks @onkwon for this PR. All looks good and I am approving this.

source/core_mqtt.c
```diff
@@ -1763,6 +1763,8 @@ static MQTTStatus_t receiveSingleIteration( MQTTContext_t * pContext,
         ( void ) memmove( pContext->networkBuffer.pBuffer,
                           &( pContext->networkBuffer.pBuffer[ totalMQTTPacketLength ] ),
                           pContext->index );
+
+        pContext->lastPacketRxTime = pContext->getTime();
```
paulbartell, line 1767:
```suggestion
        if( status == MQTTSuccess )
        {
            pContext->lastPacketRxTime = pContext->getTime();
        }
```

paulbartell: Seems prudent to update lastPacketRxTime if and only if the incoming packet was parsed successfully.

onkwon: Thanks for your suggestion. I think, it doesn't matter if the packet is a bad response or anything not expected. The only thing to be considered here is if a packet received from the broker or not since basic validation is already done above.

Am I missing something?

paulbartell: MQTT_ProcessIncomingPacketTypeAndLength checks that the id and remaining length fields are valid but does not validate the remainder of the packet. 

Since the function knows if the rest of the packet can be parsed, it might as well use that data to determine if a packet has been received.

onkwon: I got your point about the additional information. I'm still not sure if it would be really helpful though.

My whole point is if a packet received from the broker no matter what the packet is. Because, for example, even if `MQTTRecvFailed` by mismatching QoS, it should be considered received, I think. so preventing keepalive to be triggered everytime.

I'm not convinced myself either what is better. So please feel free to commit your suggestion and merge.

AniruddhaKanhere: Hello @onkwon, thanks for your response.

I am inclined to agree with @paulbartell. I think we should only update the last keep alive time if the packet is parsed correctly. Consider it "defensive programming".

Do you mind making the change or would you like us to do it for you?

onkwon: Thanks for your response. I made the change.

archigup (approved):

paulbartell (approved):

PR #236, opened 2022-12-06
RFC: Use void* for network context pointers.

paulbartell:
Specify separately from transport interface.

Using an opaque type with multiple definitions in a binary can cause a tonne of problems with debuggers. Use a void pointer to reduce the confusion here and allow transport implementations to use a more reasonable setup.


DakshitBabbar:
Hey @paulbartell,

Thanks for this draft PR. We understand that some work/testing might still be pending in this PR. For now we are to closing it. Feel free to reopen this PR if you plan to contribute to it in the near future.

Thanks again.

Issue #233, opened 2022-11-21
Keep Alive is not treated according to MQTT standard


psavyhin:
It was found that coreMQTT application that intensively sends messages to a cloud can't pass Device Advisor "Keep Alive" test.
Details are described in an attached doc.
[RD-MQTTKeepAlivetestissue-211122-1343.pdf](https://github.com/FreeRTOS/coreMQTT/files/10059365/RD-MQTTKeepAlivetestissue-211122-1343.pdf)

ademyankov:
@psavyhin I noticed some weird behavior too... I think the issue here is that the FreeRTOS repo uses very old coreMQTT version 1.1.0. When I used the latest coreMQTT v.2.1.1 I don't see any issues anymore.

FreeRTOS manifest needs to be updated for the newest version. There are some minor changes in the API so the examples will probably require change as well.

AniruddhaKanhere:
@psavyhin, this issue was resolved in PR: https://github.com/FreeRTOS/coreMQTT/pull/235

@ademyankov the FreeRTOS manifest is already updated to be beyond v2.1.1 commit. https://github.com/FreeRTOS/FreeRTOS/blob/main/manifest.yml#L41-L46. Did I miss something?


I am closing this issue. Please feel free to re-open this issue or create a new one if you think that there has been a mistake on my part.

Thank you.

PR #248, opened 2023-04-04
Fixed issue with handleKeepAlive that caused the client to constantly ping the broker

garyholiday:
### handleKeepAlive Issue
-----------

Inside of the [handleKeepAlive](https://github.com/FreeRTOS/coreMQTT/blob/main/source/core_mqtt.c#L1320) function in core_mqtt.c [on line 1356](https://github.com/FreeRTOS/coreMQTT/blob/main/source/core_mqtt.c#L1356) there is a reference to the variable `pContext->lastPacketRxTime`. This variable is supposed to be used to keep track of the last time a packet was received by the client. However, there is no further reference to this variable inside the entire project and it never gets updated.

Since this variable never gets updated, and `PINGRESP` packets do not invoke the `eventcallback` function, when the client goes to make a keep alive call, it constantly calls the broker, even though it receives an `PINGRESP` packet. This is obviously not ideal.

With the new change, when the client receives a `PINGRESP` packet, we update `pContext->lastPacketRxTime` with the latest time and now we only call the broker every `PACKET_TX_TIMEOUT_MS`.

This change could take place in several other places as well, however, since `pContext->lastPacketRxTime` is only ever used inside of the `keepAliveFunction` and it's entire purpose seems to be for monitoring pings, it seemed fine for now to only updated the received time in that use case.

Test Steps
-----------
To reproduce the original issue:
1. Turn on Debug level logging 
2. Make a connection to the broker with keep alive enabled
3. Call `MQTT_ProcessLoop` in a `while(true)` loop
4. After 30 seconds, notice the constant logs of `"Sent %ld bytes of PINGREQ packet."`

Checklist:
----------

Related Issue
-----------
https://github.com/FreeRTOS/coreMQTT/issues/233

### DIFF
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -1609,6 +1609,7 @@ static MQTTStatus_t handleIncomingAck( MQTTContext_t * pContext,
             if( ( status == MQTTSuccess ) && ( manageKeepAlive == true ) )
             {
                 pContext->waitingForPingResp = false;
+                pContext->lastPacketRxTime = pContext->getTime();
             }
 
             break;


AniruddhaKanhere:
Hello @garyholiday,

The `lastPacketRxTime` is used to note the last received packet's time so that it can be used in keep alive management as you rightly pointed out.

However, I see that the variable is correctly updated [here](https://github.com/FreeRTOS/coreMQTT/blob/19412aa6b375bdc929fd02f48b605c873e262079/source/core_mqtt.c#L1791) in the code. It should be updated why ANY packet is received, and not just the `PINGRESP` packet. Let me know if I am understanding anything incorrectly.

Thank you,
Aniruddha

pjb-idneo:
> Hello @garyholiday,
> 
> Apologies for the late response.
> 
> The `lastPacketRxTime` is used to note the last received packet's time so that it can be used in keep alive management as you rightly pointed out.
> 
> However, I see that the variable is correctly updated [here](https://github.com/FreeRTOS/coreMQTT/blob/19412aa6b375bdc929fd02f48b605c873e262079/source/core_mqtt.c#L1791) in the code. It should be updated when ANY packet is received, and not just the `PINGRESP` packet. Let me know if I am understanding anything incorrectly.
> 
> Thank you, Aniruddha

Hello @AniruddhaKanhere ,
From what I'm seeing the line that you commented (which was introduced in the following commit: d7a478aa9faef81d823c7b8b091a4c4b8a26216e) is not included in the latest tag v2.1.1, which is the commit that the esp-aws-iot module references: [https://github.com/espressif/esp-aws-iot/tree/master/libraries/coreMQTT](url). Is there any specific reason for this? Probably it's just a matter of timing but I wanted to make sure. In the meantime I will fork the esp-aws-iot repo to include the patch 

Thank you very much

AniruddhaKanhere:
> Hello @AniruddhaKanhere , From what I'm seeing the line that you commented (which was introduced in the following commit: [d7a478a](https://github.com/FreeRTOS/coreMQTT/commit/d7a478aa9faef81d823c7b8b091a4c4b8a26216e)) is not included in the latest tag v2.1.1, which is the commit that the esp-aws-iot module references: [https://github.com/espressif/esp-aws-iot/tree/master/libraries/coreMQTT](url). Is there any specific reason for this? Probably it's just a matter of timing but I wanted to make sure. In the meantime I will fork the esp-aws-iot repo to include the patch
> 
> Thank you very much

@garyholiday, the esp-aws-iot module uses the latest 'released' version of coreMQTT. But, since the last release, we have fixed a  couple of bugs. Thus, if you'd like the latest code of any repo with the fixes to all the latest set of bugs discovered since the last release, you should fetch the latest mainline commit.

> In the meantime I will fork the esp-aws-iot repo to include the patch

I think that would be the best way forward until a new version of coreMQTT is released.

Also, since the bug is fixed in the commit I mentioned, I'll be closing this PR. Feel free to open a new PR/Issue or reopen this one if you have any concern.

Regards,
Aniruddha

PR #250, opened 2023-05-03
Fix timeout calculation in sendVector function to account for overflow

AniruddhaKanhere:
Description
-----------
Currently, the `MQTTContext.getTime()` function returns a 32-bit value which represents the number of milliseconds elapsed since a given epoch. The code used a faulty comparison which did not account for overflow of the 32-bit value. This PR fixes that issue by using the predefined `calculateElapsedTime` function.

Test Steps
-----------

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -767,7 +767,7 @@ static int32_t sendMessageVector( MQTTContext_t * pContext,
                                   size_t ioVecCount )
 {
     int32_t sendResult;
-    uint32_t timeoutMs;
+    uint32_t startTime;
     TransportOutVector_t * pIoVectIterator;
     size_t vectorsToBeSent = ioVecCount;
     size_t bytesToSend = 0U;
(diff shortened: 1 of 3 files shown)


kstribrnAmzn (dismissed):

source/core_mqtt.c
```diff
@@ -832,7 +832,7 @@ static int32_t sendMessageVector( MQTTContext_t * pContext,
         }
 
         /* Check for timeout. */
-        if( pContext->getTime() >= timeoutMs )
+        if( calculateElapsedTime( pContext->getTime(), startTime ) > MQTT_SEND_TIMEOUT_MS )
```
kstribrnAmzn, line 835: What happens if `pContext` overflows and wraps? Couldn't this return a negative elapsed time then? Shouldn't that be handled here?

kstribrnAmzn: Really my core issue here is forgetting that the subtraction will ALWAYS return a positive value given the value is an unsigned integer....which is embarrassing to admit in a persisted form like a Github PR.

kstribrnAmzn (approved):

source/core_mqtt.c
AniruddhaKanhere: Making mistakes is only human :)

Thank you for the review @kstribrnAmzn.

aggarg (approved):

PR #258, opened 2023-08-23
Fix race condition in handleKeepAlive

saidrhs:
Fix race condition in handleKeepAlive

Description
-----------
The code for handleKeepAlive() (which is invoked as part of MQTT_ProcessLoop()) does not invoke the STATE_UPDATE_HOOK macros, but still reads from pContext->lastPacketTxTime. Therefore, this causes a RW race condition which is picked up by ThreadAnalyzer. With a sufficiently meddlesome scheduler, the value of lastPacketTxTime could be different for each time it is checked inside of handleKeepAlive() , causing unreliable behavior in the transmission of KeepAlive packets.

Test Steps
-----------

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -1344,6 +1344,7 @@ static MQTTStatus_t handleKeepAlive( MQTTContext_t * pContext )
     MQTTStatus_t status = MQTTSuccess;
     uint32_t now = 0U;
     uint32_t packetTxTimeoutMs = 0U;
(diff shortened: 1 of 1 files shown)


rawalexe (approved):

AniruddhaKanhere (approved):
Approved with a Nit pick about the PR description: You have said as seen from above report but I do not see any report in the description attached to the PR.

PR #276, opened 2024-02-16
Update CMakeList to optionally build UT/Coverity; Update coverity configuration

AniruddhaKanhere:
Update Coverity configuration to meet the latest coverity standard.
Updated CMakelist to only build Coverity if required instead of building the CMock based unit tests as well.

Description
-----------

Test Steps
-----------

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,7 @@ jobs:
           -G "Unix Makefiles" \
           -DCMAKE_BUILD_TYPE=Debug \
           -DBUILD_CLONE_SUBMODULES=ON \
+          -DUNITTEST=1 \
           -DCMAKE_C_FLAGS='--coverage -Wall -Wextra -Werror -DNDEBUG -DLIBRARY_LOG_LEVEL=LOG_DEBUG'
           make -C build/ all
           echo "::endgroup::"
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,5 +1,5 @@
-cmake_minimum_required ( VERSION 3.13.0 )
-project ( "CoreMQTT unit test"
+cmake_minimum_required ( VERSION 3.22.0 )
+project ( "CoreMQTT tests"
           VERSION 1.0.0
           LANGUAGES C )
 
@@ -12,6 +12,12 @@ if( NOT DEFINED CMAKE_C_STANDARD )
 endif()
 if( NOT DEFINED CMAKE_C_STANDARD_REQUIRED )
     set( CMAKE_C_STANDARD_REQUIRED ON )
+endif()
+
+# If no configuration is defined, turn everythin on.
+if( NOT DEFINED COV_ANALYSIS AND NOT DEFINED UNITTEST )
+    set( COV_ANALYSIS TRUE )
+    set( UNITTEST TRUE )
 endif()
 
 # Do not allow in-source build.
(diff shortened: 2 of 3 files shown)


test/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
-cmake_minimum_required ( VERSION 3.13.0 )
-project ( "CoreMQTT unit test"
+cmake_minimum_required ( VERSION 3.22.0 )
+project ( "CoreMQTT tests"
           VERSION 1.0.0
```
Skptak, line 3: Is this version incorrect at this point?
Know it's unrelated, but just noticed it

test/CMakeLists.txt
```diff
@@ -14,6 +14,12 @@ if( NOT DEFINED CMAKE_C_STANDARD_REQUIRED )
     set( CMAKE_C_STANDARD_REQUIRED ON )
 endif()
 
+# If no configuration is defined, turn everythin on.
```
Skptak, line 17:
```suggestion
# If no configuration is defined, turn everything on.
```

test/CMakeLists.txt
Skptak: In my opinion, I think it should be separate from the versioning of the library

The reason I think that is that I feel it provides a good way of knowing when functionally breaking changes happen in the library? I think it makes it clear that "Library <VERSION> change from <A.B.C> to <C.D.E> caused a functionality change. This caused the tests to move from <U.V.W> to <X.Y.Z>"

AniruddhaKanhere: Makes sense. I do have some different point of views though: If we were to version this same as the library, then we need not keep track of which version of tests work with which version of library. It will be the same as the library.

Additionally, since these tests are in the same repository, I don't think that anyone would choose a different/older set of tests to run against this code base. If the tests were in a separate repository, then I think properly versioning the tests make much more sense.

What do you think?

Skptak: > If we were to version this same as the library ... It will be the same as the library.

Which I would say is the single best reason to keep the versioning the same, makes it easy to understand that a version of the tests are meant for a single version of the source code and that is all we test.

> ... then we need not keep track of which version of tests work with which version of library.

And I wouldn't have much desire to maintain a matrix of what versions of the tests work with which version of the source code. But I think it can be useful for regression testing to know "These versions of the code work with these versions of the tests"

> Additionally, since these tests are in the same repository, I don't think that anyone would choose a different/older set of tests to run against this code base. 

Part of why I think they should be versioned differently is because of some of our repos that don't have the tests in them with the source code. For example the [FreeRTOS-Kernel's](https://github.com/FreeRTOS/FreeRTOS-Kernel) tests live inside of the FreeRTOS/FreeRTOS repository in the [FreeRTOS/Test](https://github.com/FreeRTOS/FreeRTOS/tree/main/FreeRTOS/Test) directory. However I recognize this isn't a huge deal, just means that a separate PR needs to be created when updating the version in the kernel repo.


But all that said, I'm more than happy to do either if somebody feels strongly about it.

.github/workflows/ci.yml
```diff
@@ -31,6 +31,7 @@ jobs:
           -G "Unix Makefiles" \
           -DCMAKE_BUILD_TYPE=Debug \
           -DBUILD_CLONE_SUBMODULES=ON \
+          -DUNITTEST=1 \
```
Skptak, line 34: Does it make sense to do coverity and unit test build in the CI separately then?

it seems like it would be really hard to make a change that breaks the unit test build but not the unit test one? But just wanted to ask about it

Skptak (approved):

n9wxu (approved):

PR #294, opened 2024-06-25
Mqttv5 connect

Pooja-Toshniwal:
MQTTV5 Connect and Connack

Description
-----------
Introduced a config variable MQTT_VERSION_5_ENABLED, which enables the users to use version 5 of MQTT. To support the new properties of connect  and connack packet introduced in the version 5, I have modified the existing data structures and functions, and added some new functions. For unit testing I have modified the existing folder structure of test to create a separate folder for v5.

Test Steps
-----------

Checklist:
----------

Related Issue
-----------


test/unit-test/MQTTv5/core_mqttv5_serializer_utest.c
```diff
@@ -0,0 +1,3299 @@
...
+    pIndexLocal[ 0 ] = UINT16_HIGH_BYTE( 13 );
+    pIndexLocal[ 1 ] = UINT16_LOW_BYTE( 13 );
+    pIndexLocal = &pIndexLocal[ 2 ];
+    status = MQTTV5_DeserializeConnack( &properties, &packetInfo, &session );
+    TEST_ASSERT_EQUAL_INT( MQTTMalformedPacket, status );
+}
+
+void test_MQTTV5_DeserializeConnackOnlyMaxQos( void )
+{
+    MQTTPacketInfo_t packetInfo;
```
amazonKamath, line 1333: I think there is duplication of the first few lines of code across the deserialize tests. Can you verify and see if duplication exists and can be wrapped in an helper function?

test/unit-test/MQTTv5/core_mqttv5_serializer_utest.c
```diff
@@ -0,0 +1,3299 @@
...
+
+    /*Validating the function Get Connect properties*/
+
+    /* Call MQTT_GetConnectPropertiesSize() with various combinations of
+     * incorrect paramters */
+
+    /*
+     * Max Packet Size cannot be null
+     */
+    properties.receiveMax = 24;
```
amazonKamath, line 2610: what about tests for other properties like session expiry? need tests for max bounds as well I guess?  I think MQTT the GetConnectPropertiesSize needs further checks for testing MAX values of sessionexpiry, receive max and maxpacket size

amazonKamath (changes requested):

test/unit-test/MQTTv5/core_mqttv5_serializer_utest.c
Pooja-Toshniwal: Max possible value of session expiry is uint32 max, receive max is uint16 max and  max packet size is 256 MB but even if  we set it beyond that it doesn't matter because it is theoretically impossible for max packet size to be more than this because of the limitation of the encoded remaining length.

test/unit-test/MQTTv5/core_mqttv5_serializer_utest.c
moninom1: The code code and initialisation for each test case can be moved to "void setUp( void )"
Which is use to setup for each test case.

moninom1: Check[ this for example](https://github.com/awslabs/amazon-kinesis-video-streams-rtp/blob/main/test/unit-test/packetization/packetization_utest.c#L23)

source/include/core_mqtt_serializer.h
```diff
@@ -96,11 +101,58 @@ typedef enum MQTTStatus
     MQTTIllegalState,     /**< An illegal state in the state record. */
     MQTTStateCollision,   /**< A collision with an existing state record entry. */
     MQTTKeepAliveTimeout, /**< Timeout while waiting for PINGRESP. */
-    MQTTNeedMoreBytes     /**< MQTT_ProcessLoop/MQTT_ReceiveLoop has received
+    MQTTNeedMoreBytes,     /**< MQTT_ProcessLoop/MQTT_ReceiveLoop has received
                           incomplete data; it should be called again (probably after
                           a delay). */
+
+    MQTTMalformedPacket=0x81,
```
moninom1, line 108: Add description here as added for above options.

test/unit-test/MQTT/core_mqtt_serializer_utest.c
```diff
@@ -2878,4 +2878,5 @@ void test_MQTT_SerializeDisconnect_Happy_Path()
     TEST_ASSERT_EQUAL( MQTTSuccess, mqttStatus );
 }
```
moninom1, line 2880: Remove this as we are not updating this file.

test/unit-test/MQTT/core_mqtt_utest.c
```diff
@@ -1116,6 +1116,7 @@ void test_MQTT_Init_Happy_Path( void )
     TEST_ASSERT_EQUAL_MEMORY( &networkBuffer, &context.networkBuffer, sizeof( networkBuffer ) );
 }
 
+
```
moninom1, line 1119: Remove all extra lines added or removed which are not needed.

source/core_mqtt_serializer.c
```diff
@@ -357,95 +588,1308 @@ static MQTTStatus_t checkPublishRemainingLength( size_t remainingLength,
...
+        else if( *pPropertyLength < sizeof( uint16_t ) )
+        {
+            status = MQTTMalformedPacket;
+        }
+        else
+        {
+            *pProperty = UINT16_DECODE( pVariableHeader );
+            pVariableHeader = &pVariableHeader[ sizeof( uint16_t ) ];
+            *pUsed = true;
+            *pPropertyLength -= sizeof( uint16_t );
```
shubnil, line 1638: Will there be a scenario where pPropertyLength can be greater than sizeof( uint16_t ). Same is true for all similar cases in the other decode functions

source/core_mqtt_serializer.c
```diff
@@ -357,95 +588,1308 @@ static MQTTStatus_t checkPublishRemainingLength( size_t remainingLength,
...
+                status = MQTTProtocolError;
+            }
+        }
+
+        *pIndex = pVariableHeader;
+        return status;
+    }
+
+    static MQTTStatus_t decodeutf_8( const char ** pProperty,
+                                     uint16_t * pLength,
```
shubnil, line 1683: Why we need the extra argument of pLength in this decode function while not in others. If we really need "pLength", then will the check in Line 1699 be against "pLength" instead of "sizeof( uint16_t )"

shubnil (changes requested):
Hi Pooja,
Thanks for the changes.
The PR looks very clean , I can see that you have taken care of most of the comments from the team review.
I have some minor comments and some clarifications.

source/core_mqtt_serializer.c
Pooja-Toshniwal: It is used to store the length of the UTF 8 string. Before decoding, property length is compared with 2 byte for validation and then length is decoded.

source/core_mqtt.c
```diff
@@ -2678,7 +3175,7 @@ MQTTStatus_t MQTT_CancelCallback( const MQTTContext_t * pContext,
 
 MQTTStatus_t MQTT_Connect( MQTTContext_t * pContext,
                            const MQTTConnectInfo_t * pConnectInfo,
-                           const MQTTPublishInfo_t * pWillInfo,
```
moninom1, line 2681: I see const is removed because we are updating pWillInfo->propertyLength, why is this specific update is needed for v5?

source/core_mqtt_serializer.c
```diff
@@ -357,95 +588,1308 @@ static MQTTStatus_t checkPublishRemainingLength( size_t remainingLength,
...
+            status = MQTTBadParameter;
+        }
+        else if( ( pAuthInfo->authDataLength != 0U ) && ( pAuthInfo->pAuthData == NULL ) )
+        {
+            status = MQTTBadParameter;
+        }
+        else
+        {
+            /*Add authentication method and data if provided*/
+            if( pAuthInfo->authMethodLength != 0U )
```
moninom1, line 1047: We have already checked authMethodLength & authDataLength != 0 above, we can remove these checks here

moninom1, line 1047: Why do we need authMethodLength & authDataLength checks again? As they are already done above. Adding extra if checks.

source/core_mqtt_serializer.c
```diff
@@ -2677,4 +4121,382 @@ MQTTStatus_t MQTT_ProcessIncomingPacketTypeAndLength( const uint8_t * pBuffer,
...
+
+    uint8_t * MQTT_SerializeConnectProperties( uint8_t * pIndex,
+                                               const MQTTConnectProperties_t * pConnectProperties )
+    {
+        uint8_t * pIndexLocal = pIndex;
+
+        pIndexLocal = encodeRemainingLength( pIndexLocal, pConnectProperties->propertyLength );
+
+        /*Serialize session expiry if provided.*/
+        if( pConnectProperties->sessionExpiry != 0U )
```
moninom1, line 4310: Multiple of these if's can be true?

source/core_mqtt_serializer.c
```diff
@@ -2677,4 +4121,382 @@ MQTTStatus_t MQTT_ProcessIncomingPacketTypeAndLength( const uint8_t * pBuffer,
...
+
+        if( pPublishInfo->willDelay != 0U )
+        {
+            *pIndexLocal = MQTT_WILL_DELAY_ID;
+            pIndexLocal++;
+            pIndexLocal[ 0 ] = UINT32_BYTE3( pPublishInfo->willDelay );
+            pIndexLocal[ 1 ] = UINT32_BYTE2( pPublishInfo->willDelay );
+            pIndexLocal[ 2 ] = UINT32_BYTE1( pPublishInfo->willDelay );
+            pIndexLocal[ 3 ] = UINT32_BYTE0( pPublishInfo->willDelay );
+            pIndexLocal = &pIndexLocal[ 4 ];
```
moninom1, line 4273: Why are we incrementing the pointer in this different way than usual? Any refernce to the existing code?

source/core_mqtt_serializer.c
```diff
@@ -2677,4 +4121,382 @@ MQTTStatus_t MQTT_ProcessIncomingPacketTypeAndLength( const uint8_t * pBuffer,
...
+
+        if( pPublishInfo->msgExpiryPresent != false )
+        {
+            *pIndexLocal = MQTT_MSG_EXPIRY_ID;
+            pIndexLocal++;
+            pIndexLocal[ 0 ] = UINT32_BYTE3( pPublishInfo->msgExpiryInterval );
+            pIndexLocal[ 1 ] = UINT32_BYTE2( pPublishInfo->msgExpiryInterval );
+            pIndexLocal[ 2 ] = UINT32_BYTE1( pPublishInfo->msgExpiryInterval );
+            pIndexLocal[ 3 ] = UINT32_BYTE0( pPublishInfo->msgExpiryInterval );
+            pIndexLocal = &pIndexLocal[ 4 ];
```
moninom1, line 4296: Why are we incrementing the pointer in this different way than usual? Any reference to the existing code?

source/core_mqtt_serializer.c
```diff
@@ -2677,4 +4121,382 @@ MQTTStatus_t MQTT_ProcessIncomingPacketTypeAndLength( const uint8_t * pBuffer,
...
+                        ( unsigned long ) *pPacketSize ) );
+        }
+
+        return status;
+    }
+
+    uint8_t * MQTT_SerializePublishProperties( const MQTTPublishInfo_t * pPublishInfo,
+                                               uint8_t * pIndex )
+    {
+        uint8_t * pIndexLocal = pIndex;
```
moninom1, line 4259: How are we ensuring pIndexLocal or pIndex has enough space to do all the writes below?

PR #305, opened 2024-09-06
MQTT Connection Status Thread Safety

DakshitBabbar:
Description
-----------

Following is a brief summary of changes:
1. Check the connected flag before doing any send operation on the connection
2. Make all the APIs that do send operations, thread safe
3. Update the connected flag within MQTT_Disconnect regardless of the return status of the send operation

Following are the specifics of the changes:
1. Add 3 new MQTTStatus_t values: MQTTStatusConnected, MQTTStatusNotConnected and MQTTStatusDisconnectPending
2. Added 1 new MQTTConnectionStatus_t value: MQTTDisconnectPending
3. Update the MQTT_Status_strerror  function to handle the new MQTTStatus_t values
4. Add a new API function MQTT_CheckConnectStatus() that will check the value of the context→connectStatus flag safely. 
5. Add this API to the core_mqtt.h file to make it available to users
6. Check the connected flag before doing any Send operation (following API's are updated)
        a. sendPublishAcks
        b. MQTT_Connect
        c. MQTT_Subscribe
        d. MQTT_Publish
        e. MQTT_Ping
        f. MQTT_Unsubscribe
        g. MQTT_Disconnect
7. Use the MQTT_PRE_STATE_UPDATE_HOOK() and MQTT_POST_STATE_UPDATE_HOOK() to make the send APIs thread safe
8. The connect status is set to MQTTDisconnectPending whenever a transport send or receive function returns a negative error code
9. `const` keyword for the the MQTTStatus_t is removed in the input parameters for the receive functions as we need to update the connection status when the receive function returns a negative error code

Relevant Explanations
---------------
- MQTT_PRE_SEND_HOOK(): The Pre and Post Send hook Macros are not required now, as the sending logic will be within the pre and post state update hook itself. (because we cannot allow other threads to change the connection state of the application until a send operation is complete).
- I have split the handleSessionResumption function. The part of that function which was handling the clean session has been added within the mutex calls in the [MQTT_Connect API](https://github.com/FreeRTOS/coreMQTT/pull/305/files#diff-2534a3c0229ae9af3801f2a5c6a24eeef2cd0a686671f0371a11d2718ba4fdd6R2828) and the unclean session part is handled by this new function that is [called outside the mutex calls](https://github.com/FreeRTOS/coreMQTT/pull/305/files#diff-2534a3c0229ae9af3801f2a5c6a24eeef2cd0a686671f0371a11d2718ba4fdd6R2866).

Pending Tasks
---------------
- [ ] Doxygen example for the new API
- [x] Unit Test Updates
- [x] CBMC Proof

### DIFF
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -2386,24 +2386,6 @@ static MQTTStatus_t receiveConnack( MQTTContext_t * pContext,
                                                       pContext->transportInterface.pNetworkContext,
                                                       pIncomingPacket );
 
-        if( status == MQTTStatusDisconnectPending )
-        {
-            /* Convert this status to MQTTRecvFailed as MQTTStatusDisconnectPending is
-             * reserved for cases where we need to let the user know about the MQTT
-             * connection status.
-             */
-            status = MQTTRecvFailed;
-
-            MQTT_PRE_STATE_UPDATE_HOOK( pContext );
-
-            if( pContext->connectStatus == MQTTConnected )
-            {
-                pContext->connectStatus = MQTTDisconnectPending;
-            }
-
-            MQTT_POST_STATE_UPDATE_HOOK( pContext );
-        }
-
         /* The loop times out based on 2 conditions.
          * 1. If timeoutMs is greater than 0:
          *    Loop times out based on the timeout calculated by getTime()
@@ -2731,7 +2713,7 @@ MQTTStatus_t MQTT_CheckConnectStatus( MQTTContext_t * pContext )
     {
         MQTT_PRE_STATE_UPDATE_HOOK( pContext );
 
-        connectStatus = pContext->connectStatus == MQTTConnected;
+        connectStatus = pContext->connectStatus;
 
         switch( connectStatus )
         {
@@ -2739,15 +2721,12 @@ MQTTStatus_t MQTT_CheckConnectStatus( MQTTContext_t * pContext )
                 status = MQTTStatusConnected;
                 break;
 
-            case MQTTNotConnected:
-                status = MQTTStatusNotConnected;
-                break;
-
             case MQTTDisconnectPending:
                 status = MQTTStatusDisconnectPending;
                 break;
 
             default:
+                status = MQTTStatusNotConnected;
                 break;
         }
 
@@ -2840,15 +2819,12 @@ MQTTStatus_t MQTT_Connect( MQTTContext_t * pContext,
                              0x00,
                              pContext->outgoingPublishRecordMaxCount * sizeof( *pContext->outgoingPublishRecords ) );
         }
-        else if( ( *pSessionPresent != true ) && ( pContext->incomingPublishRecordMaxCount > 0U ) )
+        
+        if( *pSessionPresent != true && pContext->incomingPublishRecordMaxCount > 0U )
         {
             ( void ) memset( pContext->incomingPublishRecords,
                              0x00,
                              pContext->incomingPublishRecordMaxCount * sizeof( *pContext->incomingPublishRecords ) );
-        }
-        else
-        {
-            /* MISRA Empty body */
         }
 
         pContext->connectStatus = MQTTConnected;
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -2833,22 +2833,22 @@ MQTTStatus_t MQTT_Connect( MQTTContext_t * pContext,
         pContext->index = 0;
         ( void ) memset( pContext->networkBuffer.pBuffer, 0, pContext->networkBuffer.size );
 
-        if( *pSessionPresent != true )
+        if( *pSessionPresent != true && pContext->outgoingPublishRecordMaxCount > 0U )
         {
             /* Clear any existing records if a new session is established. */
-            if( pContext->outgoingPublishRecordMaxCount > 0U )
-            {
-                ( void ) memset( pContext->outgoingPublishRecords,
-                                 0x00,
-                                 pContext->outgoingPublishRecordMaxCount * sizeof( *pContext->outgoingPublishRecords ) );
-            }
-
-            if( pContext->incomingPublishRecordMaxCount > 0U )
-            {
-                ( void ) memset( pContext->incomingPublishRecords,
-                                 0x00,
-                                 pContext->incomingPublishRecordMaxCount * sizeof( *pContext->incomingPublishRecords ) );
-            }
+            ( void ) memset( pContext->outgoingPublishRecords,
+                              0x00,
+                              pContext->outgoingPublishRecordMaxCount * sizeof( *pContext->outgoingPublishRecords ) );
+        }
+        else if( *pSessionPresent != true && pContext->incomingPublishRecordMaxCount > 0U )
+        {
+            ( void ) memset( pContext->incomingPublishRecords,
+                              0x00,
+                              pContext->incomingPublishRecordMaxCount * sizeof( *pContext->incomingPublishRecords ) );
+        }
+        else
+        {
+            /* MISRA Empty body */
         }
 
         pContext->connectStatus = MQTTConnected;
@@ -2879,7 +2879,7 @@ MQTTStatus_t MQTT_Connect( MQTTContext_t * pContext,
         LogInfo( ( "MQTT Connection is either already established or a disconnect is pending, return status = %s.",
                    MQTT_Status_strerror( status ) ) );
     }
-    else if( status == MQTTBadParameter )
+    else if( pContext == NULL )
     {
         LogError( ( "MQTT connection failed with status = %s.",
                     MQTT_Status_strerror( status ) ) );
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -444,8 +444,6 @@ static MQTTStatus_t handleUncleanSessionResumption( MQTTContext_t * pContext );
  * @brief Clears existing state records for a clean session.
  *
  * @param[in] pContext Initialized MQTT context.
- *
- * @return void.
  */
 static void handleCleanSession( MQTTContext_t * pContext );
 
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -2865,10 +2865,10 @@ MQTTStatus_t MQTT_Connect( MQTTContext_t * pContext,
         if( pContext->connectStatus == MQTTConnected )
         {
             /* This will only be executed if after the connack is received
-             * the retransmits fail for some reason on an unclean session 
+             * the retransmits fail for some reason on an unclean session
              * connection. In this case we need to retry the re-transmits
-             * which can only be done using the connect API and that can only 
-             * be done once we are disconnected, hence we ask the user to 
+             * which can only be done using the connect API and that can only
+             * be done once we are disconnected, hence we ask the user to
              * call disconnect here */
             pContext->connectStatus = MQTTDisconnectPending;
         }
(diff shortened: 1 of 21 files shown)


source/include/core_mqtt.h
```diff
@@ -414,6 +414,24 @@ MQTTStatus_t MQTT_InitStatefulQoS( MQTTContext_t * pContext,
...
+/**
+ * @brief Checks the MQTT connection status with the broker.
+ *
+ * @param[in] pContext Initialized MQTT context.
+ *
+ * @return #MQTTBadParameter if invalid parameters are passed;
+ * #MQTTAlreadyConnected if the MQTT connection is established with the broker.
+ * #MQTTDisconnected otherwise
+ *
+ * <b>Example</b>
```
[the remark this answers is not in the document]
DakshitBabbar: I am yet to write an example for this. If we do not need it then I can skip it. Please let me know.

source/include/core_mqtt_serializer.h
```diff
@@ -96,9 +96,11 @@ typedef enum MQTTStatus
     MQTTIllegalState,     /**< An illegal state in the state record. */
     MQTTStateCollision,   /**< A collision with an existing state record entry. */
     MQTTKeepAliveTimeout, /**< Timeout while waiting for PINGRESP. */
-    MQTTNeedMoreBytes     /**< MQTT_ProcessLoop/MQTT_ReceiveLoop has received
+    MQTTNeedMoreBytes,     /**< MQTT_ProcessLoop/MQTT_ReceiveLoop has received
                           incomplete data; it should be called again (probably after
                           a delay). */
+    MQTTAlreadyConnected, /**< MQTT Connection is established with the broker */
```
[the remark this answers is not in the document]
DakshitBabbar: MQTTConnected is an item of the enum MQTTConnectionStatus_t and this new name "MQTTAlreadyConnected" is supposed be an item of the enum MQTTStatus_t. So that is the reason we had to use a different name.

source/core_mqtt.c
```diff
@@ -1289,20 +1290,44 @@ static MQTTStatus_t sendPublishAcks( MQTTContext_t * pContext,
...
+            if( pContext->connectStatus != MQTTConnected)
+            {
+                status = MQTTDisconnected;
+            }
+        }
+
         if( status == MQTTSuccess )
         {
-            MQTT_PRE_SEND_HOOK( pContext );
+            // MQTT_PRE_SEND_HOOK( pContext );
```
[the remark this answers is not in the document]
DakshitBabbar: The Pre and Post Send hook Macros are not required now, as the sending logic will be within the pre and post state update hook itself. (because we cannot allow other threads to change the connection state of the application until a send operation is complete). 

And yes the commented out code will be removed, noted.

source/core_mqtt.c
```diff
@@ -1289,20 +1290,44 @@ static MQTTStatus_t sendPublishAcks( MQTTContext_t * pContext,
...
                                      localBuffer.pBuffer,
                                      MQTT_PUBLISH_ACK_PACKET_SIZE );
 
-            MQTT_POST_SEND_HOOK( pContext );
+            if( sendResult != ( int32_t ) MQTT_PUBLISH_ACK_PACKET_SIZE )
+            {
+                status = MQTTSendFailed;
+            }
+
+            // MQTT_POST_SEND_HOOK( pContext );
```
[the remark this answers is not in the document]
DakshitBabbar: noted

source/core_mqtt.c
```diff
@@ -2482,6 +2507,30 @@ static MQTTStatus_t handleSessionResumption( MQTTContext_t * pContext,
     return status;
 }
 
+static MQTTStatus_t handleUncleanSessionResumption( MQTTContext_t * pContext)
```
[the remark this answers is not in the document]
DakshitBabbar: Yes, I have split the handleSessionResumption function. The part of that function which was handling the clean session has been added within the mutex calls in the [MQTT_Connect API](https://github.com/FreeRTOS/coreMQTT/pull/305/files#diff-2534a3c0229ae9af3801f2a5c6a24eeef2cd0a686671f0371a11d2718ba4fdd6R2828) and the unclean session part is handled by this new function that is [called outside the mutex calls](https://github.com/FreeRTOS/coreMQTT/pull/305/files#diff-2534a3c0229ae9af3801f2a5c6a24eeef2cd0a686671f0371a11d2718ba4fdd6R2866). 

Please confirm if this is a good decision or not?

source/core_mqtt.c
```diff
@@ -2734,23 +2829,69 @@ MQTTStatus_t MQTT_Connect( MQTTContext_t * pContext,
...
+        /* Regardless of the status, if the mutex was taken, then it should be relinquished. */
+        MQTT_POST_STATE_UPDATE_HOOK( pContext );
+    }
+
+    if( status == MQTTSuccess && *pSessionPresent == true  )
+    {
+        /* Resend PUBRELs when reestablishing a session */
+        status = handleUncleanSessionResumption( pContext);
+    }
+
```
AniruddhaKanhere, line 2872: Should we set `pContext->connectStatus = MQTTConnected;` after the above?

Because what happens when a different thread calls `MQTT_CheckConnectStatus`? It might get a MQTTConnected even though the library has not yet handled the session resumption.

I don't think this will violate the specification (see [here](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718105)). But it seems counter intuitive that while connect has not returned a success status (and may fail during retransmits), a different function can say that the mqtt is connected.

source/core_mqtt.c
```diff
@@ -2901,7 +3067,12 @@ MQTTStatus_t MQTT_Publish( MQTTContext_t * pContext,
         }
     }
 
-    if( stateUpdateHookExecuted == true )
+    /* mutex should be released and not before updating the state
+     * because we need to make sure that the state is updated
+     * after sending the publish packet, before the receive
+     * loop receives ack for this and would want to update its state
+     */
+    if( stateUpdateHookExecuted == true )
```
AniruddhaKanhere, line 3075: I realize that this was already there, but why not fix it when we are modifying the library for the better :)

AniruddhaKanhere (commented):
Thank you for this Pull Request @DakshitBabbar!

I have left a few comments. Can you please take a look?

Also an overarching comment. We seem to be holding the mutex for a lot longer. If not absolutely necessary, please relinquish the mutex and take it back again when executing a critical section. This allows other threads to take the mutex in between the giving and taking of the mutex by the current task (think of a high priority task waiting on the mutex). In such cases, priority inversion happens, but lets try and keep it as short a duration as possible.

Thanks again!

source/core_mqtt.c
```diff
@@ -823,6 +819,11 @@ static int32_t sendMessageVector( MQTTContext_t * pContext,
         {
             bytesSentOrError = sendResult;
             LogError( ( "sendMessageVector: Unable to send packet: Network Error." ) );
+
+            if(pContext->connectStatus == MQTTConnected)
+            {
+                pContext->connectStatus = MQTTDisconnectPending;
+            }
```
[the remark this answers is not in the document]
DakshitBabbar: Correct, this function `sendMessageVector`, is called by functions ([sendConnectWithoutCopy](https://github.com/FreeRTOS/coreMQTT/pull/305/files/1e2cd7ab526b1e27dad8ec47009724a7b7dc82a4#diff-2534a3c0229ae9af3801f2a5c6a24eeef2cd0a686671f0371a11d2718ba4fdd6R2814), sendPublishWithoutCopy etc) that are guaranteed to be called after taking the mutex and before giving it back. As we cannot assume that the user implementation of MQTT_STATE_UPDATE_HOOK() will be of a recursive mutex, we should avoid taking the mutex here again.

source/core_mqtt.c
```diff
@@ -2676,6 +2715,50 @@ MQTTStatus_t MQTT_CancelCallback( const MQTTContext_t * pContext,
...
+        status = MQTTBadParameter;
+    }
+
+    if( status == MQTTSuccess )
+    {
+        MQTT_PRE_STATE_UPDATE_HOOK( pContext );
+
+        connectStatus = pContext->connectStatus == MQTTConnected;
+
+        switch (connectStatus)
```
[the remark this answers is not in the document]
DakshitBabbar: I agree with both corrections, change done.

source/core_mqtt.c
```diff
@@ -2712,14 +2797,24 @@ MQTTStatus_t MQTT_Connect( MQTTContext_t * pContext,
...
+        MQTT_PRE_STATE_UPDATE_HOOK( pContext );
+        
+        stateUpdateHookExecuted = true;
 
+        connectStatus=pContext->connectStatus;
+
+        if( connectStatus != MQTTNotConnected)
+        {
+            status = (connectStatus==MQTTConnected) ? MQTTStatusConnected: MQTTStatusDisconnectPending;
+        }
```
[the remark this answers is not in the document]
DakshitBabbar: Reasoning I applied:
If the check failed then it would be good that we will exit early, but if it passed then the Mutex will be held for a longer period of time, unnecessarily.

But yes one thing can be done:
Take the mutex, do the check in the beginning, if fails, change status and bubble it up, else relinquish the mutex and take it back when needed again. Is this what we want to do?

source/core_mqtt.c
AniruddhaKanhere: The same place to which you linked also says this: _"The Client accepts that any data that it sends before it receives a CONNACK packet from the Server will not be processed if the Server rejects the connection."_.

Is coreMQTT equipped to handle that? What happens in QoS0 packet which the client thinks are delivered to the server but are rejected because the connection was never established. I understand that QoS0 is equivalent of "fire and forget", but since underlying transport is mostly TCP/IP, a user can assume that the QoS0 packet will be delivered to the server and won't get lost. It is a _De Facto_ standard IMO (as opposed to _De Jure_).

tools/cmock/coverage.cmake
```diff
@@ -15,8 +15,9 @@ execute_process( COMMAND lcov --directory ${CMAKE_BINARY_DIR}
                          --initial
                          --capture
                          --rc lcov_branch_coverage=1
-                         --include "*source*"
                          --output-file=${CMAKE_BINARY_DIR}/base_coverage.info
+                         --include "*source*"
```
AniruddhaKanhere, line 19: Curious why this change was required. Not a blocker, I can approve with or without this change but just curious.

source/core_mqtt.c
AniruddhaKanhere: Ignore the above comment. I see that we indeed wait for the CONNACK

AniruddhaKanhere (commented):
Few minor comments. Thank you for your patience @DakshitBabbar

source/core_mqtt.c
```diff
@@ -1264,6 +1286,7 @@ static MQTTStatus_t sendPublishAcks( MQTTContext_t * pContext,
     uint8_t packetTypeByte = 0U;
     MQTTPubAckType_t packetType;
     MQTTFixedBuffer_t localBuffer;
+    MQTTConnectionStatus_t connectStatus;
```
archigup, line 1289: Why is this a parameter? Seems its just written to.

tools/cmock/coverage.cmake
DakshitBabbar: This was required because while building unit tests we were having some errors on mac. So @aggarg suggested this change along with many more to resolve them.

source/core_mqtt.c
DakshitBabbar: I am not sure what you mean exactly but it is read on line number 1313 and 1315.

source/core_mqtt.c
DakshitBabbar: fixed

AniruddhaKanhere (approved):

archigup (approved):

Issue #271, opened 2023-12-21
An issue with handling QoS2 messages when the library loses state


ademyankov:
It is a very narow corner case but stil it breaks the client and renders it unusable.

1. coreMQTT client is running on an embedded device, it subscribed to a topic with QoS2.
2. The issue happens when a message is received and then the device reboots (for unrelated to coreMQTT reasons).
3. When it boots again and re-connects to the broker it immidiately receives MQTT_PACKET_TYPE_PUBREL but since the device was rebooted the library has no knowleadge of the packet it received before, outgoingPublishRecords is empty and MQTT_ProcessLoop returns MQTTBadResponse.

**Q:** How would you suggest to handle a situation like this? There are no public functions in the library to get the failed packet id and to get access to outgoingPublishRecords structure that one could simply inject a record with id and MQTT_PACKET_TYPE_PUBCOMP type to let go that record. In the corner case like this I guess it's ok to lose the record since not much you can do if device reboots.

n9wxu:
I have restated the failure case to try and be crystal clear on what is happening.
```
sender   receiver
publish ->
         <- pubrec
** REBOOT **
pubrel  ->
           ??????
pubrel ->
pubrel ->
```

AniruddhaKanhere:
Hello @ademyankov,

Thank you for taking the time to report this bug to us.

Yes, you are absolutely correct. This bug will manifest itself when the device loses power mid QoS2 transaction.

To fix this, we believe that the state must be stored in persistent memory after a `PUBREL` is received. Whenever the device boots up, a callback would allow the application to restore the state of the MQTT 'system'. This way, when the broker sends a PUBREL back to us after reboot, the library can correctly handle it.

However, currently the library does not support it. Also, we would like if the storing and restoring the state is done by the application using callbacks from the library as the library is hardware agnostic.

We think that this can be achieved using setter and getter functions which would be called by the library whenever it receives an incoming `PUBLISH`/`PUBACK`/`PUBREL`/`PUBREC`/`PUBCOMP` and wants to store the state. This way, the application is free to choose how to implement the persistent storage so that it does not wear out same block of memory and provides fast response etc.

We also think the since this is very specific to the QoS2 publishes, we should separate the processing and handling of these packets such that anyone not using QoS2 should not have to worry about implementing the setter and getter functions.

Does the above sound like a solution to you?

[...]

PR #311, opened 2024-10-31
Add persistent storage for MQTT state

vishwamartur:
Related to #271

Add support for storing and restoring MQTT state in persistent memory to handle QoS2 messages after a device reboot.

* Add function prototypes for `MQTT_SetOutgoingPublishRecord`, `MQTT_GetOutgoingPublishRecord`, and `MQTT_GetFailedPacketId` in `source/include/core_mqtt_state.h`.
* Implement `MQTT_SetOutgoingPublishRecord`, `MQTT_GetOutgoingPublishRecord`, and `MQTT_GetFailedPacketId` functions in `source/core_mqtt_state.c`.
* Update `README.md` to include instructions on using the new setter and getter functions.
* Add unit tests for `MQTT_SetOutgoingPublishRecord`, `MQTT_GetOutgoingPublishRecord`, and `MQTT_GetFailedPacketId` functions in `test/unit-test/core_mqtt_state_utest.c`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,3 +248,38 @@ doxygen docs/doxygen/config.doxyfile
 
 See [CONTRIBUTING.md](./.github/CONTRIBUTING.md) for information on
 contributing.
+
+## Using Setter and Getter Functions for Persistent Storage
+
+The coreMQTT library provides setter and getter functions to allow the application to store and restore MQTT state in persistent memory. This is useful for handling QoS2 messages after a device reboot.
+
+### Setter Function
+
+The `MQTT_SetOutgoingPublishRecord` function allows the application to set an outgoing publish record in the MQTT context. It can be used to restore the state of the MQTT context after a device reboot.
+
+```c
+MQTTStatus_t MQTT_SetOutgoingPublishRecord( MQTTContext_t * pMqttContext,
+                                            uint16_t packetId,
+                                            MQTTQoS_t qos,
+                                            MQTTPublishState_t publishState );
+```
+
+### Getter Function
+
+The `MQTT_GetOutgoingPublishRecord` function allows the application to get an outgoing publish record from the MQTT context. It can be used to store the state of the MQTT context in persistent memory before a device reboot.
+
+```c
+MQTTStatus_t MQTT_GetOutgoingPublishRecord( const MQTTContext_t * pMqttContext,
+                                            uint16_t packetId,
+                                            MQTTQoS_t * pQos,
+                                            MQTTPublishState_t * pPublishState );
+```
+
+### Getting the Failed Packet ID
+
+The `MQTT_GetFailedPacketId` function allows the application to get the packet ID of the failed packet from the MQTT context. It can be used to handle the situation when the library loses state after a device reboot.
+
+```c
+MQTTStatus_t MQTT_GetFailedPacketId( const MQTTContext_t * pMqttContext,
+                                     uint16_t * pPacketId );
+```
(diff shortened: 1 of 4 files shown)


archigup:
Hey, thanks for opening a PR!

We recently landed having the library handle QoS1/2 re-transmits for the user to the main branch; this seems like its related. We should consider how these are used together

The device could set its packet storage callbacks to use flash, though when initializing the library need to restore the in-transmit state as you have here.

AniruddhaKanhere:
Hello @vishwamartur!

Thank you for raising this PR. It seems quite useful.

Let us take a look. We recently have added functionality for the library to be able to re-send publishes which were not ACKed before a disconnect and reconnect of MQTT connection with an unclean session. We will see whether this might be somehow linked with that feature.

Thanks again for taking the time to create the PR!

\- Aniruddha

README.md
```diff
@@ -248,3 +248,38 @@ doxygen docs/doxygen/config.doxyfile
...
 See [CONTRIBUTING.md](./.github/CONTRIBUTING.md) for information on
 contributing.
+
+## Using Setter and Getter Functions for Persistent Storage
+
+The coreMQTT library provides setter and getter functions to allow the application to store and restore MQTT state in persistent memory. This is useful for handling QoS2 messages after a device reboot.
+
+### Setter Function
+
+The `MQTT_SetOutgoingPublishRecord` function allows the application to set an outgoing publish record in the MQTT context. It can be used to restore the state of the MQTT context after a device reboot.
```
AniruddhaKanhere, line 258: I wonder whether a callback might be more suitable here?

Here is what i am thinking: The method described here will work perfectly fine when the reboot is planned. But what if the reboot is unplanned such as a power failure or link to the internet goes down.

In such cases, application might not have enough time to ping the library repeatedly for new packets. I understand that there might be race condition even in callback. But it is less pronounced.

When the application calls the API described above, it can either do that very frequently leading to higher CPU consumption, or it can do that slowly leading to a higher chance of a missed QoS2 publish when a reboot happens.

vishwamartur:
Hi @AniruddhaKanhere and @archigup,

Thank you for the feedback! I'm glad you see the potential of this feature.

Regarding the callback suggestion, that's an interesting point. A callback approach would indeed reduce the risk of missing QoS2 messages, particularly in the case of unexpected reboots or network issues, where the application might not have enough time to poll for unacknowledged packets.

I’ll explore implementing a callback mechanism to handle persistent storage updates, as this might provide a more robust solution for unplanned reboots. Additionally, I could look into ways to make the polling frequency adjustable to balance CPU usage and the likelihood of packet loss. Let me know if you have any specific requirements or additional insights on this.

Thank you again for the thorough review and suggestions!

Best regards,
Vishwa

README.md
AniruddhaKanhere: I thought more about it and I would like to fix my error in the above statement.

In case of unplanned power loss/reboot, the coreMQTT library cannot be held responsible for keeping the actual state and the stored state in NVM in sync. Because regardless of whatever the library does, there will always be a race condition in the actual and stored state of the system. 

Thus, I think the library should only support planned reboots. In case of unplanned reboots, I think the application knows/understands the system best and it should try to fix any issues with the state of the system.

I will think more on this. But let me know if you have any suggestions or concerns @vishwamartur.

Thanks

DakshitBabbar:
Hey @vishwamartur,
The solution posted in [the conversation thread of the Issue #271](https://github.com/FreeRTOS/coreMQTT/issues/271#issuecomment-2581848724) is the preferred approach . Let us know if you have any concerns regarding that.

Thanks.

DakshitBabbar:
Closing this PR. Feel free to re-open if you have any concerns.

Issue #317, opened 2025-04-03
Express the need of a non-blocking TransportRecv_t


diopoex:
I think the documentation could be made more clearly on these two cases:

![Image](https://github.com/user-attachments/assets/8f3b3445-7386-4c9a-a976-6a97096c02fb)
https://freertos.github.io/coreMQTT/v2.3.1/group__mqtt__callback__types.html#ga227df31d6daf07e5d833537c12130167

I think there could directly be mentioned that if the transport Interface is blocking the keepAlive mechanism isnt working, because usually blocking sockets only return 0 on recv() when the peer has closed the connection(see: https://github.com/FreeRTOS/coreMQTT/blob/8ec72d6ef7d61c78d899f069395ac5fed7624650/source/core_mqtt.c#L1769).

Then a mention in the "Porting Guide" and/or in the "Transport Interface" could also make sense.

diopoex:
If this makes sense I will gladly make the changes and create a pull request.

rohitmadan07:
Thanks for your contribution ! The PR has been merged. Closing this issue

PR #318, opened 2025-04-04
Express the need of a non-blocking TransportRecv_t

diopoex:
Express the need of a non-blocking TransportRecv_t #317

Description
-----------
- changed documentation to express the need of a non-blocking TransportRecv_t 
- mentioned that a non-blocking implementation is highly recommended in porting.dox and in the transportinterface
- did go more into detail in TransportRecv_t what the effects of a blocking recv implementation are

Test Steps
-----------
only changed documentation

Checklist:
----------

Related Issue
-----------
#317 

### DIFF
--- a/source/interface/transport_interface.h
+++ b/source/interface/transport_interface.h
@@ -200,9 +200,9 @@ typedef struct NetworkContext NetworkContext_t;
  * implementation does NOT block.
  * coreMQTT will continue to call the transport interface if it receives
  * a partial packet until it accumulates enough data to get the complete
- * MQTT packet. 
- * A non‐blocking implementation is also essential so that the library's inbuilt 
- * keep‐alive mechanism can work properly, given the user chooses to use 
+ * MQTT packet.
+ * A non‐blocking implementation is also essential so that the library's inbuilt
+ * keep‐alive mechanism can work properly, given the user chooses to use
  * that over their own keep alive mechanism.
  *
  * @param[in] pNetworkContext Implementation-defined network context.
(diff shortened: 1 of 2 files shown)


source/interface/transport_interface.h
```diff
@@ -197,6 +198,8 @@ typedef struct NetworkContext NetworkContext_t;
  *
  * @note It is HIGHLY RECOMMENDED that the transport receive
  * implementation does NOT block.
+ * A non‐blocking implementation is essential so that the keep‐alive
+ * mechanism can work properly.
  * coreMQTT will continue to call the transport interface if it receives
  * a partial packet until it accumulates enough data to get the complete
  * MQTT packet.
```
DakshitBabbar, line 205: I was wondering if it would be better to include the fact that the user can also implement their own keep alive mechanism in which case the non-blocking receive function may/may not affect the keep alive mechanism. Something like this:
```c
/*
 * @note It is HIGHLY RECOMMENDED that the transport receive
 * implementation does NOT block.
 * coreMQTT will continue to call the transport interface if it receives
 * a partial packet until it accumulates enough data to get the complete
 * MQTT packet. 
 * A non‐blocking implementation is also essential so that the library's inbuilt 
 * keep‐alive mechanism can work properly, given the user chooses to use 
 * that over their own keep alive mechanism.
 */
 ```
Let me know what you think.

DakshitBabbar (commented):
Hey @diopoex,
Thanks for this PR, looks good to me. I've added a small suggestion, let me know your thoughts.

DakshitBabbar (approved):

rohitmadan07 (approved):

Issue #321, opened 2025-05-20
Unit tests should fail but do not due to lack of assert


jasonpcarroll:
A number of the unit tests are not doing what they are supposed to for coreMQTT. This is because the unit tests are compiled without assert, which causes the underlying coreMQTT library to exhibit behavior that is not correct and should cause some of our unit tests to fail / assert out, but they do not as there is no assert... for example this test https://github.com/FreeRTOS/coreMQTT/blob/42d843f40ef2abc752a419f45f7adb7dac67e591/test/unit-test/core_mqtt_utest.c#L1545 does not set writev to NULL (which it should as it is trying to test with transport send sending no bytes, not writev). Normally, this test would hit this assert because the writev implementation is used and not the send implementation as it should https://github.com/FreeRTOS/coreMQTT/blob/42d843f40ef2abc752a419f45f7adb7dac67e591/source/core_mqtt.c#L828-L829

DakshitBabbar:
The PR #322 has been merged.

Closing this issue.
Thanks.

PR #322, opened 2025-05-28
Update unit tests to work with asserts enabled

DakshitBabbar:
Description
-----------
This PR solves the Issue #321. 

**Learning:** Unit tests are supposed to be built with asserts enabled during the development process. But when these are built for coverage analysis these should be built with asserts disabled. This is because the lcov tool would consider all the branches in the asserts as uncovered. 

This PR:
1. Update the unit tests to work with asserts enabled
2. Update the code to have 100% branch coverage. To achieve this, one branch which was logically unreachable, was excluded from the coverage analysis.
3. Updates the CI checks to build the code with asserts to check if all the unit tests are working as expected. Along with this it will also build the code without asserts to check coverage.
4. Updates the command in the README.md that users should use to build the unit tests. This command has asserts disabled, as asserts are only required during the development process of the unit tests. 

Test Steps
-----------
The unit tests are building and running successfully with asserts. They also give 100% coverage without asserts. 

Checklist:
----------

Related Issue
-----------
#321

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
         uses: actions/checkout@v3
 
       - env:
-          stepName: Build CoreMQTT
+          stepName: Build CoreMQTT with asserts
         run: |
           # ${{ env.stepName }}
           echo -e "::group::${{ env.bashInfo }} ${{ env.stepName }} ${{ env.bashEnd }}"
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -32,7 +32,29 @@ jobs:
...
+          echo -e "::group::${{ env.bashInfo }} ${{ env.stepName }} ${{ env.bashEnd }}"
+
+          sudo apt-get install -y lcov
+          cmake -S test -B build/ \
+          -G "Unix Makefiles" \
+          -DCMAKE_BUILD_TYPE=Debug \
+          -DBUILD_CLONE_SUBMODULES=ON \
+          -DUNITTEST=1 \
+          -DCOV_ANALYSIS=1 \
+          -DCMAKE_C_FLAGS='--coverage -Wall -Wextra -Wsign-compare -Werror -DLIBRARY_LOG_LEVEL=LOG_DEBUG'
```
aggarg, line 57: The step says without assert but there is no `NDEBUG` flag here. Is that correct?

DakshitBabbar: The `-DCOV_ANALYSIS=1` flag is the one which disables the asserts in case of coverage analysis build. So I thought it would be better to use that instead of `-DNDEBUG`.

aggarg: Perfect, thanks!

aggarg (approved):

kar-rahul-aws (approved):

PR #356, opened 2026-02-18
Fix timeout check ordering in sendMessageVector

benjacam:
Fix timeout check ordering in sendMessageVector

Description
-----------
I am running your library with MQTT_SEND_TIMEOUT_MS=0. My intention is to not block at all when attempting a send. 
The library is running in a RTOS thread, and the getTimeFunction is returning the RTOS tick timer value (which is updated in another thread). In this scenario, the writev transport function successfully completed the requested send. But during the send, the other thread incremented the timer once. This means, that when this line executes:
`if( calculateElapsedTime( pContext->getTime(), startTime ) > MQTT_SEND_TIMEOUT_MS )`
the code considers the timeout elapsed, and breaks from the while loop, _before_ advancing pIoVectIterator and decrementing vectorsToBeSent.  This means that the packet was sent, but it isn't cleared from the pending vectors.

The fix is to move the calculateElapsedTime to the very end of the while loop, after the vect state is advanced.

Test Steps
-----------
Set MQTT_SEND_TIMEOUT_MS 0. Observe logs: "sendMessageVector: Unable to send packet: Timed out." , but also add to this log the values of bytesSentOrError and bytesToSend. Eventually, you will see a log where the timeout is reported, but bytesSentOrError ==bytesToSend. Or add assertion in the body of the if( calculateElapsedTime.

Checklist:
----------

- [ x] I have tested my changes. No regression in existing tests.

Related Issue
-----------

### DIFF
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -875,8 +875,10 @@ static int32_t sendMessageVector( MQTTContext_t * pContext,
             pIoVectIterator->iov_len -= ( size_t ) sendResult;
         }
 
-                /* Check for timeout. */
-        if( calculateElapsedTime( pContext->getTime(), startTime ) > MQTT_SEND_TIMEOUT_MS )
+        /* Check for timeout. */
+        if( ( bytesSentOrError < ( int32_t ) bytesToSend ) &&
+            ( bytesSentOrError >= 0 ) &&
+            ( calculateElapsedTime( pContext->getTime(), startTime ) > MQTT_SEND_TIMEOUT_MS ) )
         {
             LogError( ( "sendMessageVector: Unable to send packet: Timed out." ) );
             break;
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -851,13 +851,6 @@ static int32_t sendMessageVector( MQTTContext_t * pContext,
             /* MISRA Empty body */
         }
 
-        /* Check for timeout. */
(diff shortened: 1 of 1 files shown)


AniruddhaKanhere (changes requested):
There is a minor update which I would like to propose. Can you please take a look @benjacam?

AniruddhaKanhere:
@benjacam, we have since merged in MQTTv5 changes which do fix this issue. That is why now we see a merge conflict.

Apologies that I could not get your PR merged in before the MQTTv5 changes went in - I intend to make amends. To that end, I am planning to cut a branch off of the old library code and merge in your fix. Then we can release that library as well for anyone who wants to continue using MQTTv3.

I do urge you @benjacam and @blocksebastian to try out MQTTv5 which was released recently: https://github.com/FreeRTOS/coreMQTT/releases/tag/v5.0.0 and let us know how can we make it better and more useful. You will find the porting guide here if you are interested: https://github.com/FreeRTOS/coreMQTT/blob/v5.0.0/MQTTv5Guide.md

AniruddhaKanhere:
@benjacam can you edit this PR to be committing to https://github.com/FreeRTOS/coreMQTT/tree/v3-CodeBase instead of main? That way we can get this change into the MQTTv3 code base and release that for others to use.

aadityasg:
@benjacam can you please create this PR on the branch @AniruddhaKanhere mentioned earlier?

PR #376, opened 2026-07-01
Documentation and cleanup fixes

AniruddhaKanhere:
Description
-----------

Test Steps
-----------

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -2907,7 +2907,7 @@ static MQTTStatus_t sendUnsubscribeWithoutCopy( MQTTContext_t * pContext,
 
         if( CHECK_SIZE_T_OVERFLOWS_32BIT( pIterator->iov_len ) ||
             ADDITION_WILL_OVERFLOW_U32( totalPacketLength, ( uint32_t ) pIterator->iov_len ) ||
-            ( ( totalPacketLength + pIterator->iov_len ) > MQTT_MAX_REMAINING_LENGTH ) )
+            ( ( totalPacketLength + pIterator->iov_len ) > MQTT_MAX_PACKET_SIZE ) )
         {
             LogError( ( "Total packet length must less than 268435461." ) );
             status = MQTTBadParameter;
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -4276,8 +4276,10 @@ MQTTStatus_t MQTT_InitStatefulQoS( MQTTContext_t * pContext,
     else
     {
         pContext->incomingPublishRecordMaxCount = incomingPublishCount;
+        pContext->incomingPublishRecordMaxCountConfigured = incomingPublishCount;
         pContext->incomingPublishRecords = pIncomingPublishRecords;
         pContext->outgoingPublishRecordMaxCount = outgoingPublishCount;
+        pContext->outgoingPublishRecordMaxCountConfigured = outgoingPublishCount;
         pContext->outgoingPublishRecords = pOutgoingPublishRecords;
 
         if( ( pAckPropsBuf != NULL ) && ( ackPropsBufLength != 0U ) )
@@ -4556,6 +4558,12 @@ MQTTStatus_t MQTT_Connect( MQTTContext_t * pContext,
          **/
         if( status == MQTTSuccess )
         {
+            /* Start from the application-configured limits so the effective
+             * limits are recomputed fresh on every connect, rather than being
+             * monotonically shrunk across reconnects. */
+            pContext->incomingPublishRecordMaxCount = pContext->incomingPublishRecordMaxCountConfigured;
+            pContext->outgoingPublishRecordMaxCount = pContext->outgoingPublishRecordMaxCountConfigured;
+
             if( pContext->connectionProperties.receiveMax < pContext->incomingPublishRecordMaxCount )
             {
                 pContext->incomingPublishRecordMaxCount = pContext->connectionProperties.receiveMax;
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -4727,6 +4727,7 @@ MQTTStatus_t MQTT_Publish( MQTTContext_t * pContext,
     MQTTPublishState_t publishStatus = MQTTStateNull;
     MQTTConnectionStatus_t connectStatus;
     uint16_t topicAlias = 0U;
+    bool stateReserved = false;
 
     /* Maximum number of bytes required by the 'fixed' part of the PUBLISH
      * packet header according to the MQTT specifications.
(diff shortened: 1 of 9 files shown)


archigup (dismissed):

source/include/core_mqtt.h
```diff
@@ -1425,7 +1442,10 @@ MQTTStatus_t MQTT_MatchTopic( const char * pTopicName,
  *  - 0xA1 - Subscription identifiers not supported.
  *  - 0xA2 - Wildcard subscriptions not supported.
  *
- * Refer to #MQTTSubAckStatus_t for the status codes.
+ * The codes above are the raw reason-code bytes returned in the SUBACK payload.
+ * #MQTTSubAckStatus_t enumerates only the common subset (success at QoS 0/1/2 and
+ * the generic failure code 0x80); the remaining reason codes are returned as their
+ * raw byte values without a corresponding enumerator.
```
kstribrnAmzn: This comment is confusing me - [the return type](https://github.com/AniruddhaKanhere/coreMQTT/blob/main/source/include/core_mqtt_serializer.h#L307-L332) doesn't account for the raw reason codes. How are you returning them from this function?

source/core_mqtt.c
```diff
@@ -2907,7 +2907,7 @@ static MQTTStatus_t sendUnsubscribeWithoutCopy( MQTTContext_t * pContext,
 
         if( CHECK_SIZE_T_OVERFLOWS_32BIT( pIterator->iov_len ) ||
             ADDITION_WILL_OVERFLOW_U32( totalPacketLength, ( uint32_t ) pIterator->iov_len ) ||
-            ( ( totalPacketLength + pIterator->iov_len ) > MQTT_MAX_REMAINING_LENGTH ) )
+            ( ( totalPacketLength + pIterator->iov_len ) > MQTT_MAX_PACKET_SIZE ) )
```
kstribrnAmzn, line 2910: Why did this need to switch away from max remaining length?

test/unit-test/core_mqtt_serializer_utest.c
```diff
@@ -3088,6 +3088,95 @@ void test_MQTTV5_suback( void )
...
+        0x00              /* Payload: Reason code = Success */
+    };
+
+    memset( &properties, 0x00, sizeof( properties ) );
+    properties.maxPacketSize = MQTT_MAX_PACKET_SIZE;
+
+    subackPacket.type = MQTT_PACKET_TYPE_SUBACK;
+    subackPacket.remainingLength = 10;
+    subackPacket.headerLength = 2;
+    subackPacket.pRemainingData = &packetBuffer[ 2 ];
```
kstribrnAmzn, line 3125: Why not simply remove the first two fields and do `&packetBuffer`?

source/core_mqtt.c
AniruddhaKanhere: `totalPacketLength` is the full packet size, not the Remaining Length field. The first vector pushed here is the fixed+variable header (totalPacketLength += pIterator->iov_len), so by this branch it already counts the header bytes. The other two checks in this function (the per-topic-filter check and the trailing totalPacketLength > MQTT_MAX_PACKET_SIZE) and sendPublishWithoutCopy already use MQTT_MAX_PACKET_SIZE; the property-buffer branch was the inconsistent one and was ~5 bytes too strict, so this makes it correct and consistent.

source/core_mqtt.c
```diff
@@ -4812,6 +4827,26 @@ MQTTStatus_t MQTT_Publish( MQTTContext_t * pContext,
...
+                 * packet was not sent. Remove it so the outgoing publish record
+                 * slot is not leaked over repeated send failures. The original
+                 * send-failure status is preserved for the caller. Records that
+                 * already existed (duplicate retransmits) are left intact. */
+                removeStatus = MQTT_RemoveStateRecord( pContext, packetId );
+
+                /* The record was just reserved by this call for a QoS1/QoS2
+                 * publish, so it is guaranteed to be present and removable. A
+                 * failure here can only mean the state records were corrupted. */
+                assert( removeStatus == MQTTSuccess );
```
kstribrnAmzn, line 4845: This checks out to me but how would a state record even have existed for a duplicate message in first place (https://amzn-aws.slack.com/archives/D02RQLN93PC/p1783030359889739)?

test/unit-test/core_mqtt_serializer_utest.c
AniruddhaKanhere: `pRemainingData` must point at the variable header (packet id onward), i.e. after the 2-byte fixed header. packetBuffer begins with the fixed header (0x90 + the remaining-length byte), so `&packetBuffer[2]` skips it. Passing `&packetBuffer` would feed those two header bytes into the parser as the packet id / property length and mis-parse. This mirrors the real receive path, where `MQTTPacketInfo_t.pRemainingData` points to the bytes after the fixed header (as produced by `MQTT_GetIncomingPacketTypeAndLength`). `remainingLength` is also required: `MQTT_DeserializeAck` parses exactly remainingLength bytes from `pRemainingData`, so leaving it `0` would parse nothing.

`headerLength` is the one field that is genuinely unused by `MQTT_DeserializeAck` here (I set it only to mirror a realistic packet and the sibling `test_MQTTV5_suback`). I'm happy to drop the headerLength line if you prefer the minimal setup.

source/core_mqtt.c
AniruddhaKanhere: If application decides to send a publish again before it receives an ACK, it is supposed to set the dup flag.

source/include/core_mqtt.h
AniruddhaKanhere: Fixed - made it more clear

source/core_mqtt.c
kstribrnAmzn: Thank you for the explanation here

test/unit-test/core_mqtt_serializer_utest.c
kstribrnAmzn: As we discussed - I'm okay with leaving them as long as the corresponding comment mentions that they are unused and left only to mirror a realistic packet.

test/unit-test/core_mqtt_serializer_utest.c
```diff
@@ -3088,6 +3088,95 @@ void test_MQTTV5_suback( void )
...
+    MQTTReasonCodeInfo_t subackReasonCodes;
+    MQTTPropBuilder_t propBuffer = { 0 };
+    size_t propIndex = 0;
+    uint8_t propertyId = 0;
+
+    /* SUBACK carrying one property (Reason String "abc") and one reason code. */
+    uint8_t packetBuffer[ 12 ] =
+    {
+        0x90,             /* Fixed header: SUBACK type */
+        0x0A,             /* Remaining Length = 10 */
```
kstribrnAmzn:
```suggestion
        0x90,             /* [Unused, mirrors real packet] Fixed header: SUBACK type */
        0x0A,             /* [Unused, mirrors real packet] Remaining Length = 10 */
```

kstribrnAmzn (approved):

rawalexe (approved):

